PR #1, opened 2025-04-09
Add OEIS A034258 results up to N=200

uhrm:
This commit adds results for extending the OEIS sequence A034258 up to N=200 using the linear programming approach for factoring factorials.

For each number in the sequence, there are two result files
`FC_{N}_{T}.txt`
`FC_{N}_{T+1}.txt`
The first file establishes a factorization of N! into N factors larger than or equal to T. The second file establishes an upper bound for the number of factors of N! with a threshold T+1 that is strictly smaller than N.

The approach fails for N=155 since the LP upper bound for N=155, T=46 is 155 (exactly), but the method can only find 154 factors.


uhrm:
With this pull request, I've added only the results. I'm happy to also add the code for reproducing these results to this repository, if desired.

uhrm:
I've noticed that results up to N=200 are already known, so I've extended the calculations up to N=600. Luckily, there are no corner-cases between 200 and 600.

teorth:
Thanks!  If you are willing to share the code, that would also be great (plenty of room in the repository).

If I understand correctly, we do not currently have a factorization that witnesses the value $t(156)=46$ that is claimed in the OEIS data?  It might be nice to use some ad hoc method to locate such a factorization, just to complete the dataset.

uhrm:
The problem is with $`N=155`$. We *do* have factorization that witnesses $`t(155) = 45`$, but the linear programming upper bound cannot rule out $`t(155) = 46`$.

I noticed that when solving the optimization problem as an integer program (i.e. requiring the variables to only take integer values), Gurobi is able to presolve more aggressively and immediately establish $`t(155) < 46`$ without adding any cutting planes or doing branch and bound search.

teorth:
Ah, thanks for clarifying! I've modified the description of results on the github page accordingly.

PR #2, opened 2025-04-10
Add linear programming code

uhrm:
Add the linear programming code.

Additionally, this pull request updates the OEIS result files:
- I changed the file format to include the threshold in the first line of the result files to make it more self-contained and validation a bit easier. This is compatible with the committed version of the code.
- There was a bug in the code where it didn't compute a numerically valid certificate when the upper bound was tight. (This doesn't change any conclusions - the OEIS values are correct.)


teorth:
It occurs to me that this code could be used to probe the lower limits of Conjecture 2, by demonstrating moderately large values of $N$ where one can prove that $t(N) < N/3$.  Currently this threshold is known (contingent on Conjecture 2 being true for $N>10^9$) to be in the range $(600, 8 \times 10^4)$ and there is a weak heuristic prediction that it is somewhere around $7 \times 10^3$.

uhrm:
The code can certainly handle problem instances of these sizes. I will start a run to try and pinpoint the exact value (or at least narrow the interval significantly).

plby:
@uhrm : Can you verify that t(43631) < 14544 ?

uhrm:
I solved all instances down from $`N=8*10^4`$ with $`t(N)=N/3`$ using the LP solver. The first instance that failed was $`N=43631`$ ($`t(43631)=14544`$) where it could only find $`N-1`$ factors. The LP upper bound is 43630.962609465037, thus confirming the value of @plby.

I have a bunch of factorization files laying around :-). Do you think it's worth committing them to the repository?

plby:
@uhrm : Great, thank you! 

In my separate pull request https://github.com/teorth/erdos-guy-selfridge/pull/5 I had tried adding some factorization files but in a different format and "sparsely". In my commit message here https://github.com/teorth/erdos-guy-selfridge/pull/5/commits/8933353a178001839bfb672502dc021d2247e065 I describe at the end how you only need some values.

The goal was indeed to link up with the n>=80,000 results. I'm glad you are able to produce all of these factorizations.

teorth:
That's excellent news!  Perhaps the factorization files near 43631 can be committed, but the full list of files does not need to be as long as the code is uploaded, and is relatively quick to run.

There may also be some sporadic values of $N < 43631$ for which $t(N) \geq N/3$, but perhaps the LP method can rule this out for all but a small handful of cases.

plby:
@uhrm : Perhaps you could commit a file with just the value of $t(n)$ for all $n$ for which it is known to you, up to $8\cdot 10^4$?

uhrm:
@plby: The values $`t(N)`$ are not readily available from this run. I solved the problems keeping $`t=\lceil N/3 \rceil`$ fixed and optimizing the number of factors. This is the 'natural' mode of operation for the LP formulation I'm using. Calculating $`t(N)`$ would require an additional outer loop for doing bisection. This was not necessary for verifying the $`N/3`$ lower bound.

@teorth: If there is interest in finding also smaller $`N`$ for which $`t(N) \ge N/3`$ (or ruling it out), I could extend the search. In fact, I will do it anyways.

PR #4, opened 2025-04-11
Greedy Algorithm Speedups

Straw17:
Implemented some small speedups for the greedy algorithm. In particular:
- Added the ability to toggle the assert statements at the end. It turned out that multiplying all the factors together was taking around 10x - 20x times as long as the actual algorithm runtime
- When looking at the factors, I noticed that the same numbers were being factored out multiple times (often hundreds of times). I changed the loop to automatically factor out a number as many times as it can in one go rather than running multiple loops. This yielded a 2x - 3x speedup.
- Changed from having our primes as a dictionary to having them as a list. This did not yield a major speedup, but I hope it will come in handy with some of the further optimizations I've been thinking about.

### DIFF
--- a/src/python/greedy.py
+++ b/src/python/greedy.py
@@ -5,7 +5,7 @@ import time
 
 # A greedy algorithm to factorize, adapted from Maple code of Andrew Sutherland. (Thanks to Adam Wagner for help with the code conversion to Python.)
 
-def solve(N,T):
+def solve(N, T, skip_tests=True):
     P = list(sympy.primerange(1, T))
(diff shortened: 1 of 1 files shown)


Straw17:
It is worth noting that we spend the vast majority of the loops on factoring out large primes that are paired with small numbers. For example, for N=30000 and T=10000, we spend the first ~4500 loops on factoring out primes that are paired with two (dividing out 2 * 99991, then 2 * 99989, and so on). A similar pattern occurs for 3, 4, and so on. I am trying to think of ways to efficiently do these updates as batches (dividing out 2^k multiplied by the k largest primes, for example), since that seems like the best way to get a large speedup.

PR #5, opened 2025-04-11
Add factorizations that prove Conjecture 2 in the range [43632, 44760].

plby:
The first commit below adds a simple program simple_check.py that verifies putative factorizations with a very simple logic and a very simple file format.  The second commit below adds factorizations that prove t(n) >= n/3 for n in the range [43623, 44760].  It takes approximately two minutes to verify this all.

The reason for the bottom of the range is that I believe the threshold 43632 is optimal.  The reason for the top of the range is that I simply have not yet extended it further.


plby:
Upon reflection, I don't know whether it's the best idea to add such large files to the repository.  Combined, they are 28M -- and I had originally planned to add more..

However, they compress quite well: together they are just 1.2M.

teorth:
Thanks for this!  Can you provide some details about the method you used to obtain the factorizations?

As regards file size, perhaps a compressed version would be reasonable.  A possible alternative is to report the factorization with multiplicity: for instance 6 6 6 6 6 5 5 5 5 might be reported instead as 6 5 \\ 5 4, though this may require significant refactoring of the code (and I'm not sure how much compression in the file size this would create).

plby:
I'm using the same linear programming setup everyone else is, except I wrote my own quick implementation.  Also, I don't have gurobi so I'm using the open-source solver https://en.wikipedia.org/wiki/Lp_solve instead.

I had originally avoided using multiplicity to keep simple_check.py super simple, but I also hadn't computed the filesizes yet.

I've completed the "ladder" of factorizations up to 80,000, to be added in my next commit.

plby:
I have added a commit that increases the range to [43632, 80973], in particular linking up with previous work.

Verifying all the files using simple_check.py took me ~17 minutes.

The total file size is now 95M.  If instead we compress all the files, the size would be 4.1M.  I could create a separate commit and PR with that if we want to avoid bulking up the repository too much.  (Or explore multiplicity or other compression, etc.)

plby:
By the way, note that @uhrm has verified the tightness of the threshold over here: https://github.com/teorth/erdos-guy-selfridge/pull/2#issuecomment-2796186186

Straw17:
@plby If you have a chance, would you mind adding your linear programming code? It'd be nice to have a non-Gurobi version around.

plby:
@teorth: This commit already includes only select values of $N$. I tried including as sparse a collection as possible that proves the whole range. I wrote more about it in my detailed commit message here: https://github.com/teorth/erdos-guy-selfridge/pull/5/commits/8933353a178001839bfb672502dc021d2247e065

They are closer together for smaller $n$ (just $3$ apart at first) but the last two are $567$ apart because that's how much slack is available.

I think the modular condition requires focusing on $1\bmod 3$ actually, because we want $t(3k+1) \ge k+1/3$. Accordingly, those are what I included (except for the threshold $43632$ itself, which is a multiple of $3$).

teorth:
Ah, okay, thanks for the clarification.  Perhaps one can then just compress all the files except for the 43632 one, which is perhaps of the most interest?

plby:
Sounds good, done.  This PR adds a program simple_check.py that verifies a factorization.  It can transparently accept a compressed file, so there is no need to manually decompress anything.

This PR also adds factorizations for 319 sparsely-chosen n in the range [43632, 80973].  All but the first are gzip-compressed; the factorization of the threshold n=43632 is not.

I have checked all of the factorizations, and I have checked that they cover the stated range when taking into account the obvious addition of n+1 to a factorization of n.

The total file size is under 5M now.  The git history of these commits does *not* include commits with the uncompressed data, because I force-pushed them away.

PR #8, opened 2025-04-12
Investigating Conjecture 2 in Lower Ranges

Straw17:
In https://github.com/teorth/erdos-guy-selfridge/pull/2#issuecomment-2797900971, @teorth discussed the fact that $t(N) \geq N/3$ for some sporadic numbers beneath the boundary. In order to investigate this, I added some functionality to easily do mass checks using the existing linear programming code:

- Added a mass_check file to repeatedly run the main function from the linear programming code, and report the results
- Altered the linear programming code to report if the rigorous upper bound is less than $N$ (report failure)

Using that code, I checked the ranges $N \in [42000, 43631]$ and $N \in [200, 15000]$. I first checked the 42000 - 43631 range to see how many successes / failures we have around the crossover point. The 200 - 15000 range is because I was curious if we would have any crossovers in very early numbers, though I knew it was quite unlikely (there were no successes in that range). I've added this data as two files.


teorth:
Thanks!  Ideally we would also run the lower bound construction to make sure that the successes really are successes, but given how close the upper and lower bounds track, I would imagine that most of them are.

It should now be possible to determine the smallest $N$ for which Conjecture 2 holds (presumably a little lower than 42000), by a combination of the linear programming upper and lower bounds.

Straw17:
I set up a run through the range $N \in [40000, 42000)$ this morning, and it found one success at $41006$. Unless there are other successes for $N < 40000$, I believe that should be smallest $N$ for which Conjecture 2 holds. I'll put in another run for the $N \in [38000, 40000)$ range, since no successes there as well would be strong heuristic evidence in favor of $41006$ as the smallest.

teorth:
Just to clarify, this is just a run of the upper bound at present?  So for instance it is not yet confirmed that $t(N) \geq N/3$ holds at $N=41006$, only that the linear program upper bound fails to disprove this.

Straw17:
Correct! My terminology was a bit unclear. I was attempting to run the linear program on all the inputs within the range to check to see if the upper bound established by the linear programming solution was strictly less than $N$. This allows easy identification of which $N$ the conjecture cannot hold for, but does not say if the conjecture *does* hold for a particular $N$, as it only yields an upper bound.

My plan was to use this process to identity potential candidates for the conjecture holding, and then do a second pass to attempt to factorize those candidates to obtain a confirmation.

Straw17:
Also, some updates:

- My run for $N \in [38000, 40000) completed, and confirmed conjecture 2 cannot hold for any $N$ in this range. As I previously noted, I personally think this is decent empirical evidence for $N = 41006$ as the smallest $N$ where conjecture 2 holds, though it would be nice to have more rigorous confirmation of this.
- I confirmed that conjecture 2 does in fact hold for $N = 41006$. I've attached the factorization file ([FC_41006.txt](https://github.com/user-attachments/files/19721551/FC_41006.txt)) to this comment for anyone who would like to reference it, and I can also commit it to the repository if there is any interest.

teorth:
Thanks for the update!  Sure, this is worth putting in the repository as well, given that it is the likely minimum value of $N$ for Conjecture 2.

PR #13, opened 2025-04-13
Add a new proof that Conjecture 2 holds for sufficiently large N.

plby:
I think I have a different proof that Conjecture 2 holds for sufficiently large N.  The result is aided by a computation, which I have included in this PR.

The proof is constructive, operating in three phases we call "down", "reorganize", and "up".  (The directory name is thus `dnup` meaning "down-up".)  We start with the usual factorization of $N! = 1\cdot 2\cdot 3\cdot \dotsb\cdot N$.  We will move around factors so that each of the $N$ terms is greater than $N/3$.

The algorithm has some parameters: $K_{dn}$ and $K_{up}$, as well as a detailed table of numbers of length $K_{up}$.  We have chosen $K_{dn}=K_{up}=2^{11} = 2048$.  It is not necessary that they be equal, nor powers of two.  However, a small step later is slightly easier if $K_{up}$ is a power of two, and so we have chosen them to both equal the smallest power of two that seems to work; call the exponent $V=11$.

## Down

First the down step.  We begin with the list of factors $L_1=(0,N]$.  We will split it up into $K_{dn}$ different lists by dividing by the primes up to $K_{dn}$.  We iterate over the primes in order:

For the prime $2$, we initially split $L_1$ up into its odd and even terms and we divide each even term by $2$.  We thus obtain $N/2$ factors of $2$ and two lists.  The first list (which we will continue to call $L_1$ as we modify it) consists of the odd numbers in $(0,N]$ while the second list $L_2$ has all numbers in $(0,N/2]$.  We continue by splitting off $(0,N/2]$ into $(0,N/4]$ and so forth.  After we're done with the prime $2$, we will have $L_1$ being the odd half of $(0, N]$, $L_2$ is the odd half of $(0, N/2]$, and so on, until finally $L_V$ is all of $(0,N/2^V]$.  We have also collected a bunch of powers of $2$.

Then we consider the prime $3$.  The first step is to take $L_1$ and remove the multiples of $3$.  $L_1$ is then left with one-third of the numbers in $(0,N]$ (namely, those congruent to $1$ or $5\bmod 6$) while $L_3$ becomes half of the numbers in $(0,N/3]$; it is only half because recall that $L_1$ was already missing the even numbers.  Also, we get $N/6$ multiple of $3$ as a result.  And we continue for the rest of the uses of $3$ as well as the other primes.

After the end of the down step, we will have collected a "budget" of primes up to $K_{dn}$ as well as $K_{dn}$ lists $L_1, \dotsc, L_{K_{dn}}$ where each $L_k$ consists of some fraction of $(0,N/k]$.

## Reorganize

Now we reorganize the lists by grouping them based on the least multiple they need to get up to $N/3$.  For example, $L_1$ consists of some fraction of $(0,N]$.  We re-group it into sublists $S_1$ a fraction of $(N/3,N]$, $S_2$ a fraction of $(N/6,N/3]$, $S_3$ in $(N/9, N/6]$.  This is because the first sublist $S_1$ already exceeds $N/3$; the second sublist should be multiplied by two; the third $S_3$, by three; and so on.  We stop this process at $K_{up}$, so the final sublist $K_{up}$ is part of $(\frac{N/3}{K_{up}}, \frac{N/3}{K_{up}-1}]$.  We also have one special sublist $S_{*}$ as part of $(0,\frac{N/3}{K_{up}}]$, which will require various multiples strictly and arbitrarily larger than $K_{up}$.

When we next consider $L_2$, which is some fraction of $(0,N/2]$, we will take its part in $(N/3,N/2]$ and add it to the existing sublist $S_1$.  That is, we forget which $L_k$ something came from, now simply tracking how many factors end up in each $S_k$.

## Up

Now we have a bunch of sublists $S_1, \dotsc, S_{K_{up}}$ indexed by what multiple they need to get back up to $N/3$.  There's also a special sublist $S_{*}$.  We will use out prime budgets from the down step to multiply back up.

Let's discuss the special sublist $S_{*}$ first, which recall is a bunch of numbers less than $\frac{N/3}{K_{up}}$.  We will fix it entirely using powers of $2$.  Half of the list needs $V+1$ powers of two, a quarter of the list needs $V+2$ powers of two, and so forth.  Adding this up, on average they need $V+2$ powers of two.

The other sublists are handled using the "detailed table of numbers of length $K_{up}$" mentioned earlier.  This table says how much of each $S_k$ to multiply by $k$, and how much to multiply by a larger number.  More specifically, it is just a list of numbers $m_1, \dotsc, m_{K_{up}}$ where each $m_k$ says what fraction of the length of the original list (that is, $N$) to multiply by $k$.  The multiples are just done in order: So $m_1$ of $S_1$ will be multiplied by $1$, then $m_2$ of whatever remains of $S_1$ as well as the next part $S_2$ will be multiplied by $2$, and so forth.

For this step to succeed, we need to make sure we don't overspend our budget.  That is precisely what the program in this pull request does.

## Computation

The program included here computes that for the given parameters of the algorithm above ($K_{dn}=K_{up}=2^{11}$ and a big table of numbers), the algorithm succeeds.  This is all done with **exact arithmetic** working with fractions of $N$.  The program has a reasonably large amount of comments, and it *attempts* to be fairly straightforward.

The inequalities involved (like not over-spending the budget) have all been made to have **slack**.  This is because of course when you work with an actual value of $N$, the numbers like $N/2$ and $N/3$ are not exactly correct.  The program reports the amount of slack available.  When figuring out the sublist sizes that need to be multiplied up, there's always more than $10^{-6} N$ extra room available.  Similarly, there's always more than $10^{-6} N$ powers of $p$ available after spending each prime budget.

I have furthermore arranged for the remaining prime budget of $2$ to be larger, namely more than $10^{-4} N$.

All in all, this means that this algorithm should succeed for sufficiently large $N$.  But unfortunately, I don't know how large!


teorth:
Very nice!  As I understand it, what you are doing is taking the numbers $n$ from $1$ to $N$, greedily pulling out small primes from $n$ and stopping before one goes above $K_{dn}$, achieving a factorization $n = d m$ where $d \leq K_{dn}$ and $m$ obeys certain congruence conditions.  Then, provided $m \geq N/K_{up}$, one replaces $n$ with $m \lceil t/m \rceil$, using the prime factors saved from removing $d$ to redistribute to $\lceil t/m \rceil$, with a final ad hoc step to use powers of two to deal with the case when $m < N/K_{up}$.

There may be a version of this argument that could be easier to study analytically, and get a civilized threshold beyond which Conjecture 2 holds.  Pick some fixed threshold $K$.  For any number $n$ between $1$ to $N$, split it into an $K$-smooth number $d$ (all prime factors at most $K$) and a $K$-rough number $m$ (all prime factors at least $K$).  For large $N$, the $K$-rough numbers have density $\prod_{p \leq K} (1-1/p)$ (about $e^{-\gamma} /\log K$, by Mertens' theorem), and each $m$ appears $\{ d \leq N/m: d \hbox{ is } K\hbox{-smooth}\}$ times, so asymptotically one has a precise count on how many $m$'s one gets in any interval of the form $[t/k, t/(k-1)]$.  For the roundup step, one has to multiply each $m$ that shows up in the interval $[t/k, t/(k-1)]$ not by $k$, but by the first $K$-smooth number greater than or equal to $k$ (this will be $k$ itself for all $k \leq K$, and for many $k$ beyond that, but after that one has to make some skips).

I am hoping that with a reasonable choice of $K$ (similar to your $2^{11}$ choice), one can show that this algorithm stays under budget to handle all cases other than when $m$ is in some tiny interval like your $[0,N/K_{up}]$, and then do something ad hoc such as the powers of 2 method to handle that remaining interval.  The point is to analyze the error terms when $N$ is large, one only needs to understand the distribution of $K$-rough numbers in a given interval, which is something which is very well studied and for which there should be quite good estimates from sieve theory.

plby:
From the big picture, my goal here was (and is) indeed to get a "civilized threshold" for Conjecture 2.  I haven't managed yet but I'm still hopeful.  In particular, I think I can implement this algorithm to take a specific $N$ and then see how it does empirically.

I agree with your reframing in terms of factorizations.  One of the reasons I was looking at it the way I was is that I was using that the numbers are evenly-distributed within their intervals; I guess I can see that from your description too.  This is needed particularly for the special interval: I can't have too many small numbers there or else the multiples get too big.

However, there is one detail in your version that is different from what I'm doing: I don't replace $n$ with $m\lceil t/m\rceil$, but rather $m u$ for some multiple $u\ge \lceil t/m\rceil$ given by the table of numbers included.  I kind of sped over this point in my original writeup, but I multiply the largest $N m_1$ numbers by $1$, then the next $N m_2$ by $2$, then the next $N m_3$ by $3$ and so forth, but it is not necessarily the case that I multiply all the numbers that need only a factor of $\lceil t/m\rceil$ by that exactly: I often multiply by a larger factor.

This turns out to be important for my analysis.  For example, $\lceil t/m\rceil$ is at most $K_{dn}/3$; but I extracted some primes larger than $K_{dn}/3$ and I want/need to use them!  So my table computes how to redistribute the primes/multiples effectively while staying under budget.  In fact, I think that's where the juice comes from, so much so that I think doing it the naive way won't work for **any** cutoff.

Regarding pulling out all primes: I was indeed previously trying to pull out all factors of $p$, getting the easily-computed budget of $N/(p-1)$, but I was having trouble analyzing how much re-upping was needed later if you pulled out a lot of powers of $p$, which is why my method stops at $K_{dn}$ right now.

plby:
I got myself a little confused: $\lceil t/m\rceil$ can be bigger than I said.  I remain concerned about using enough of the primes bigger than $K_{dn}/3$, but I guess not for the reason I said.

Nonetheless, I do not just use the smallest possible multiple $\lceil t/m\rceil$.  For example, I save up multiples of $2$ so I can use them later.  If you look at https://github.com/teorth/erdos-guy-selfridge/blob/main/src/dnup/data.py you can see it jumps back and forth and in particular is 0 for a lot of numbers including (but not limited to) larger powers of $2$.

teorth:
Thanks for clarifying the role of the multiplier table data.  How was this table generated?  By a linear program?

How much slack is there, in the sense of ratio between the budgeted amount of primes and the spent amount of primes (for the worst choice of prime)?  If there is not a lot of slack then one may need to take $N$ to be enormous in order for sieve theory to give accurate enough control on all the sieved sets here.  (For $N$ a multiple of 2048! everything should be OK, although some care needs to be taken for the special intervals, but of course this is too huge to qualify as a "civilized" bound.)

teorth:
I looked more into the sieve theory and unfortunately the convergence rates to the asymptotic density is not great.  The main problem is the fluctuation in the Buchstab function https://en.wikipedia.org/wiki/Buchstab_function that controls the density of the numbers up to x that are not divisible by any prime up to x^{1/u}.  The practical upshot of this is that if one wants to use this sort of approach for a realistically sized N, say 10^12, then the densities here are only reliable if one is using quite small primes, e.g., onl using numbers with prime factors at most 11 or so.  If there was some way to reorganize the argument to rely almost entirely on small primes then there would be a chance of a decent non-asymptotic result.

plby:
Yes, the table was generated by a linear program.  There are decision variables $m_1, \dotsc, m_{2048}$ for how much to multiply up.  They are subject to "earth-moving" constraints
1. $m_1\le \lvert S_1\rvert/N,$
2. $m_1+m_2\le (\lvert S_1\rvert+\lvert S_2\rvert)/N$,
3. $m_1+m_2+m_3\le (\lvert S_1\rvert+\lvert S_2\rvert+\lvert S_3\rvert)/N$,
and so on, which says you can't "take credit" for multiplying too much back up.  I also insist that everything is taken care of separately from the special set: $\sum_k m_k = 1-\lvert S_{*}\rvert/N$.  (This is reasonable anyway because that means everything is taken care of by $2048$, which is a power of $2$, the same prime used for the special set.)

There are also constraints on the prime budgets.  For example, for $p=2$, the amount spent $m_2+\dotsb+m_{2048}$ (plus the special set amount!) is bounded above by the $2$ budget (which is a little less than $1$).  For $p=3$, $m_3+\dotsb+m_{2046}$ is bounded by the $3$ budget (slightly under $1/2$), and so forth.

For all of the inequality constraints, I insist on there being some slack available.  This is for two reasons: one is that the linear program solver computes with floating point, so when I later try to verify with exact arithmetic, I need room; the other reason is of course that I wanted some kind of asymptotic bound where slack is needed as well.

The program computes (exactly) the additive amount of slack, which is approximately $1.4\cdot 10^{-6}$.  By that I mean in every inequality, the smaller side is at least that much smaller than the bigger side.  I was trying to make this slack large, and it can't go that much smaller because the amount of budget available for large primes is very small.

You asked about the *multiplicative* slack for the prime budgets.  It is approximately $2.9\cdot 10^{-6}$ for this witness.  (Because of how I was optimizing, that's just the additive slack divided by the $3$ budget.)  If I try to increase the multiplicative slack, I can get it slightly over $10^{-4}$.  (In addition, there is $3\cdot 10^{-3}$ extra available from the $2$ budget, which I thought could be useful in the bounds somehow.)

PR #14, opened 2025-04-14
Adding the C++ code that I used to verify the conjecture up to 10^11

matthieu-rosenfeld:
Here is the C++ code that I used.

I am hesitant to put the ouput file, as it dot not contain much information.  Obviously, outputing the factorization for each N! is too large even if we just give it in the form "factor #number of occurrence of this factor". I was thinking of ignoring the numbers that contain large prime factors (say up to n/9 since what we do for these primes is uniquely determined by pi(n) for n, n/3, n/6 and n/9), the output would be much smaller, but it still is really large (for 97483479 we still use 742853 factors, so I assume it would take around a GB for integers of order 10^11) and we need to give it for a few hundred values. Any suggestion on what could be a meaningful output would be interesting.

### DIFF
--- a/src/from_medium_to_large/RangeOfValues.cpp
+++ b/src/from_medium_to_large/RangeOfValues.cpp
@@ -18,7 +18,7 @@ cc_hash_table<long long, long long> largestDivisor;
 // these constant define the interval for which we run the program
 // we require minN*minN >= maxN
 #define minN 10000000LL
-#define maxN 100000000LL
+#define maxN 100000000000LL
 
 // this is the number of consecutive intervals that are used for precomputing primes and the factorization
 #define NUMBEROFINTERVALS 400
@@ -209,7 +209,6 @@ long long bestFact(long long n, long long eps = 0){
   long long i = findIndexLargerPrime(n);
   for(; primes[i]>n; i--){}
   for(; primes[i]>=targetVal; i--){
-    long long p = primes[i];
     factorizer.addToFactorization(factorizer.factors[i], 1, i);
   }
   long long prevGoal = 2;
--- a/src/from_medium_to_large/RangeOfValues.cpp
+++ b/src/from_medium_to_large/RangeOfValues.cpp
@@ -1,0 +1,266 @@
+#include <iostream>
+#include <vector>
+#include <unordered_map>
+
+using namespace std;
+
+//the following might not be available on your machine (it should on a linux machine whit compiler g++)
+//in this case simply replace the cc_hash_table by unordered_map which are slower
+#include <ext/pb_ds/assoc_container.hpp>
+using namespace __gnu_pbds;
+cc_hash_table<long long, long long> smallestDivisor;
+cc_hash_table<long long, long long> largestDivisor;
+
+// unordered_map<long long, long long> smallestDivisor;
+// unordered_map<long long, long long> largestDivisor;
+
+
+// these constant define the interval for which we run the program
+// we require minN*minN >= maxN
+#define minN 10000000LL
+#define maxN 100000000LL
+
+// this is the number of consecutive intervals that are used for precomputing primes and the factorization
+#define NUMBEROFINTERVALS 400
+
+vector<long long> primes;
+
+
+///////////////////////////////////
+//// Precomputations of primes ////
+///////////////////////////////////
+
+//compute the primes numbers and factorization for interval [0, e]
+void factorizationOfFirstInterval(long long e){
+  vector<long long> firstDivisor(e+1, -1);
+  for(long long i=2; i<firstDivisor.size(); i++){
+    if(firstDivisor[i] != -1){
+      long long fd = primes[firstDivisor[i]];
+      largestDivisor[i] = max(fd, largestDivisor[i/fd]);
+    } else {
+      primes.push_back(i);
+      largestDivisor[i] = i;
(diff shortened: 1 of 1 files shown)


teorth:
Hmm, yeah, a gigabyte is perhaps a bit much, and if these  files can be generated in a few minutes from the code for each input that may well be sufficient.  Perhaps one thing that would be interesting is a histogram of factor counts in various bins (e.g., rounding down to the nearest multiple of a thousand) to get a sense of how efficient the factorization really is, say for the specific N that are powers of 10 (10^7, 10^8, 10^9, 10^10, 10^11).   For these specific values of N one could also report on the surplus at N/3, and (if feasible) what the best lower bound on t(N) is for these values (in particular, how it compares with the asymptotic prediction of N/e - c_0 N / log N).   Perhaps one could report on the surplus for a somewhat denser set of sampled N, enough to get a reasonable graph of what the surplus looks like as a function of N (I assume it grows somewhat linearly, but with a small positive slope and many random-looking irregularities).

kevinventullo:
@matthieu-rosenfeld, based on @teorth 's comment [here](https://terrytao.wordpress.com/2025/03/26/decomposing-a-factorial-into-large-factors/#comment-687766), the full conjecture will be resolved if we can verify up to 10^12. You mentioned this code took 15h w/ 218GB of RAM to verify 10^11, and that running time is sublinear. So it seems that one could verify up to 10^12 in under a week, although the memory may grow.

Are you currently running this and waiting for it to finish, or do you need some more tricks to get the memory usage down?

evan-conway:
I just made a comment about this (and optimization suggestions for accomplishing it) on the blog, though it's waiting for moderator approval. It might be worth opening a dedicated issue thread here to discuss the push for $10^{12}$ and related optimization possibilities. If I'm understanding correctly, it seems like memory and the running time of the prime sieve are the biggest problems here, so I've mainly been thinking about that. I've attached my post below for reference.

image

matthieu-rosenfeld:
@kevinventullo currently the code is not running. There is a machine with 600GB of RAM that I can easily use, but it will not be enough with the current implementation. There is a 3TB machine that I could probably use and that should be enough, but it is more complicated for me to access it. I will start investigating on how to use it on Monday, in case there is no other solution.

@evan-conway I will answer your comment from the blog here:

I agree, the precomputation is the slowest part of the code, and storing the primes and the factorizations is what occupies a lot of memory. Although for $n =10^{12}$ there are $38\times 10^9$ prime numbers below $n$, so simply keeping the factorization of $n!$ will occupy a lot of memory (320GB if we use 64bits integers, we could divide that in half by using 32 bits for all primes that appears less than 2^32 so starting at ~1009). One solution mentioned in the blog by @teorth is to only compute the factorization exactly for primes up to say $n/9$ and to use a good enough approximation of the prime counting function $\pi$ for the other ones (since all primes between $n/9$ and $n/8$ are treated the same and have the same multiplicity in n!, and it should be fine if we lose only a small fraction of them). Instead of using some known approximation for $\pi$, we could also simply do the sieve, but for large $n$ we could store only the value of $\pi(n)$ for every n divisible by $10000$ (that should give a much better approximation than the approximation that have error of order $\sqrt{n}$ if I understand correctly). There is still the problem of sieving and storing the factorizations... The best solution for now would be to store only the smallest divisor of a number when it is larger than say 13 which should decrease the size of this set by a multiplicative factor.

* The code uses a hashmap, because I am not keeping the factorization of all numbers. I only keep the "possibly" useful numbers (where the criterion to decide whether we $f$ could be useful is the one I mentioned in the blog, ie  $\frac{f\times P_+(f)}{P_-(f)}\le n$ where $P_+$ and $P_-$ are respectively the largest and smallest prime divisor). This eliminates a big fraction of integers below $n$ (if I remember correctly it seems to be of order $ \log n$, it is probably not too hard to estimate this asymptotically). However, the hashmap is still big because the memory usage per integer stored is way less efficient (there might be a better choice than cc_hash_map...). I tried using a sorted array, but a binary search is much slower than using a hashmap... I have been thinking of using a sorted array with some extra information and relying on the fact that we usually need numbers that are closed to the one we used before, so we could do better than a binary search, but it is not completely clear to me... An easy way forward would be to find a way to eliminate even more numbers. 

* The sieve is currently a segmented sieve done block by block, but these are large blocks (currently in 400 blocks, this is the parameter NUMBEROFINTERVALS). I will try to reduce the block size to see if it improves things.

* I suppose that relying on an external efficient sieving tool could be a reasonable solution, but it does not solve the factorization problem. I will try to look if there is some idea that I can easily implement in the code.

* Currently, the code is not storing the first divisor of a number divisible by $2$(except for the first block, but it should not make a big difference). It is true that using the same simple trick on the sieving could save up to a factor two in run time ! 

* I understand how you could jump to the next multiple of $1024$ (which should not increase the number of factorizations that we need to compute, as it is really negligible). But I do not understand how to use the fact that $N$ is a multiple of $1024$ ? I guess it could be more useful to know that the target $t$ is divisible by $1024$, but I do not know what to do with this either. Could you elaborate ?

* Well... I will definitely add the -march=native flag (I used the O3).

kevinventullo:
@matthieu-rosenfeld I've added a PR here for skipping small primes and computing them on the fly: https://github.com/teorth/erdos-guy-selfridge/pull/28

PR #16, opened 2025-04-14
Algorithm to find last failures

evan-conway:
As part of this project, we were interested in finding the last $N$ such that conjecture 2, $T(N) \geq N/3$, failed to hold. However, we did not seem to have a good method for automatically conjecturing which point would be the last failure, or even finding the general range.

I was thinking about ways to automatically search for these last failure points, and came up with the following algorithm:

1. Start with a large range of numbers (for example $N \in [10^1, 10^6]$ where you think it is possible for your last failure point to be
2. Generate random points in this range according to a log-uniform distribution (so that you're randomly sampling across various scales) and then add both of the endpoints (to make sure the full range is covered)
3. Evaluate each of your points, noting if they seem to obey the conjecture
4. Based on your evaluations, choose a new, smaller range to repeat the algorithm on (move the lower bound up to the highest failure you observed, move the upper bound down to some point based on how many successes you have after that point)

One of the main issues with this approach is that you have no guarantee of correctly finding the last failure point. This is because you can numerically find lower bounds on the last failure point (simply by finding a higher point where conjecture 2 fails to hold), but you do not have a good way to show that there cannot be any failures after a given point. Due to this, I decided that if we observe a certain number of consecutive successes in the random points we sample, then we assume that there will be no more failures after that point.

In practice, this algorithm seems to work decently. For example, testing with $T = N/3$ correctly gives 43631 as the last failure. I also ran the algorithm on a few other inputs of the form $T = N \cdot k$, in order to automatically come up with some conjectures:

- The algorithm conjectures that $T = N \cdot 0.3$ holds for all $N > 391$
- The algorithm conjectures that $T = N \cdot 0.31$ holds for all $N > 984$
- The algorithm conjectures that $T = N \cdot 0.32$ holds for all $N > 3126$
- The algorithm conjectures that $T = N \cdot 0.33$ holds for all $N > 19058$

As a small bonus, I believe this algorithm can be easily adapted to search for the first instance where the conjecture does hold.

### DIFF
--- a/src/linprog/facfac.py
+++ b/src/linprog/facfac.py
@@ -329,7 +329,12 @@ def main(N: int, T: int, rigorous: bool = False, ilp: bool = False, save: str = 
     if save:
         write_factorization(N, T, fact, save)
 
-    return meets_bound # return whether the upper bound is >= N
+    # If the greedy algorithm returns a valid factorization, use that. Otherwise, fall back to checking the linear programming upper bound.
+    if sum(fact[1].values()) >= N:
+        meets_bound = True
+
+    # return whether this seems to meet the requirements
+    return meets_bound
 
 
 
--- a/src/linprog/find_crossover.py
+++ b/src/linprog/find_crossover.py
@@ -1,0 +1,61 @@
+from facfac import main
+import numpy as np
+import math
+
+output = open(f"crossover_output.txt", "w")
+
+# generate a random number from the log uniform distribution between two bounds
+def log_random(lower, upper):
+    lower_log = math.log(lower)
+    upper_log = math.log(upper)
+    random_log = np.random.uniform(lower_log, upper_log)
+    random = math.exp(random_log)
+    return math.floor(random) # convert to integer
+
+LOW = 10**1 # low end of intitial range
+HIGH = 10**6 # high end of initial range
+SAMPLE_COUNT = 50 # number of samples to take
+CUTOFF = 25 # number of successes to cut off after
+
+def f(x):
+    return main(x, math.ceil(x/3)) # change T here
+
+def last_failure(lower, upper):
+    output.write(f"Searching range [{lower}, {upper}]\n")
+
+    # if the range is small enough to just examine the entire range, do that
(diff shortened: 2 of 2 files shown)


Straw17:
To clarify, when you say "upper bounds", are you referring to upper bounds on the last point where the conjecture fails to hold, or something else? One of the weaknesses of this approach is that it doesn't actually give any upper bounds on the last point where the conjecture fails to hold, only "heuristic" bounds where we observe many successes after a certain point.

With regard to the idea of using the crossover points to predict other crossover points (if that is indeed what you are proposing), the main issue is that this algorithm is limited to a fairly small range. Since the linear programming approach does not scale well beyond around $N = 10^6$ (based on my observations), this approach would work quite poorly if you wished to find the crossover point for something like $T(N) \geq T \cdot 0.36$. My hope is that Matthieu's greedy algorithm could help with this issue by helping extend to higher ranges (though that has the issue of only being able to confirm that the conjecture does hold for a specific $N$, and cannot confirm that the conjecture does not hold for that $N$). This constraint means that I have a fairly limited number of data points, so I'm unsure how well extrapolation would work.

teorth:
There is a dual linear program that can be used to give upper bounds; see Lemma 3.1 of https://github.com/teorth/erdos-guy-selfridge/blob/main/LaTeX/notes.pdf .  For instance in https://github.com/teorth/erdos-guy-selfridge/pull/17 Boris showed that to factor 43630! into factors exceeding 43630/3 one can only have at most 43629.962... factors, so Conjecture 2 fails at this value.

Straw17:
Ah, I see! I am currently using that upper bound with linear programming to detect when $N$ fails conjecture 2 (if not, I use the greedy algorithm to check if it succeeds). I now understand what you were asking with regard to the lower and upper bounds. I'm currently doing a run to check the remaining values in the range $N \in [15000, 38000]$ in an attempt to wrap up that bit of conjecture 2, but when that finishes in a couple of hours I can start another run to get your desired plots.

evan-conway:
Closing this pull request since all the contents are in my newer pull request.

PR #22, opened 2025-04-15
Crossover Points + Bounds on t(N)

evan-conway:
This pull request contains the code for my previous crossover points pull request (https://github.com/teorth/erdos-guy-selfridge/pull/16), along with some extra code to graph out bounds of $t(N)/N$.

In https://github.com/teorth/erdos-guy-selfridge/pull/16#issuecomment-2802318197, @teorth had expressed an interest in seeing a plot of upper and lower bounds on $t(N)$. I used the same method as @uhrm's OEIS code to obtain increasingly tight upper and lower bounds on $t(N)$ for randomly chosen points in $N \in [10^2, 10^5]$. This yielded the following graph of the behavior of $t(N)/N$, which I found to be pretty interesting
output

I've included the precise data used to make this chart in Data/t_large_n_bounds.txt, if anyone wants to examine it.


teorth:
Interesting!  I imagine that the piecewise constant behavior at large N (and the consistent gap between the upper and lower bounds) is an artefact of the thresholds t being tested, that should presumably disappear if one did more testing?

evan-conway:
I was also surprised by that behavior, and need to investigate it further. One thing to note is that the upper and lower bounds were made with two different methods (linear programming for the upper and the greedy algorithm for the lower), so it seems surprising that they would both jump at the same time, unless $t(N)/N$ itself had a jump. I would be curious to hear if it seems plausible for $t(N)/N$ to actually behave in such a way for larger $N$. My thought was perhaps that the jumps could be due to some specific $N$ being added that makes it particularly easy to factor in the desired way, but that's pure conjecture.

The gap between the lines is a function of how I structured things. @uhrm's algorithm, which I used, works by essentially doing a binary search for $t(N)$, choosing a potential value and then seeing if it produces a valid factorization. This means that the gap can be approximately halved by doing one more iteration. I chose five iterations, but that was purely because it seemed to offer a good tradeoff between having acceptable speed and having a decently tight bound.

Since I'd like to figure out the cause of the jumps, my next plan is to zoom in on one of the discontinuities and see how exactly the jump occurs (ideally figuring out the exact $N$ where the jump happens). I'll likely look at the one just before $10^4$, since that seems relatively easy to analyze. I'll report back when I have more results.

evan-conway:
Testing with a larger number of iterations on the binary search (moving the bounds to make them more precise) seems to yield results that are more in line with expectations:
tn_exact

It seems like $t(N)/N$ becomes somewhat less spiky for larger $N$, which I suppose makes sense.

teorth:
That is interesting!  I wonder how it compares against the predicted value of $1/e - c_0/\log N$.  It seems a lot of the irregularities in factoring at small N get smoothed out at large N.  One reason for this, I think, is that it is easier to "make change": if one wants to reach a certain target size t and one already has some partial factorization, it becomes increasingly easier to use small primes like 2 and 3 to come very close to the target t, when t is large (the fact that log 2 and log 3 are incommensurable allows one to play more and more games at larger scales).

evan-conway:
In the range shown, it seems to be quite a bit below $1/e - c_0 / \log(N)$
tn_exact_line
It seems to be getting closer towards the upper ranges, so there is noticeable convergence.

I also realized that one could rearrange $y = 1/e - (c_0 + o(1)) / \log(N)$ to get a sense of what the $o(1)$ term looks like. Rearranging in this manner to get $o(1) = (1/e - y) \cdot \log(N) - c_0$ (somewhat of an abuse of notation, but you get the idea) gives a graph of the $o(1)$ term as something like
output2
It is not terribly clear to me what this will look like for higher $N$.

teorth:
Thanks for this!  Very tentatively it looks like there may be a second order term in the asymptotics, thus $t(N) \approx 1/e - c_0/\log N - c_1/\log^2 N$ for some constant $c_1$ (which may be somewhere around 0.7, but the data is too scant to make a really firm prediction here).

matthieu-rosenfeld:
My run with $n$ up to $10^{11}$ outputted for each number $n$ for which it ran, the value $\varepsilon$ for which it succeeded. The way $\varepsilon$ is used, means $t(n)>(n(1000+\varepsilon))/3000$ (this is up to rounding as is the plot and everything in the rest of this message). So I can provide data to continue this plot up to $10^{11}$. Here is the plot:

output

Because of the way I did it $\varepsilon$ was chosen in $\\{60,50,40,30,...\\}$, so the result is really constant piecewise. The more granular lower bound is obtained by rerunning it with more granularity for $\varepsilon$ up to $10^9$. It is slower (by a factor 4 for the range $[10^7,10^9]$), but I could rerun it for $n$ up to $10^{11}$ (or a bit further if, we need to go a bit further). Note that the formal lower bound implied over this range by these points is $t(n)\ge n/3$, as whenever we prove $t(n)>(n+x)/3$, then the next $n$ for which we do something is $n+x+1$ (see Kevin Ventullo's comment on the blog post).


Whenever this pull request is accepted, I could merge the data, or should we keep them separate since they are not generated with the same tools ?


In my previous pull request, @teorth also mentioned plotting the number of factors larger than $n/3$. I did not do that since it is not what the code computes (the surplus is useful to jump from $n$ to $(1+\varepsilon)n$, while having extra factors does not help much).

evan-conway:
@matthieu-rosenfeld do you want to just make a separate pull request? I'll probably be adding another commit or two to this due to @teorth's request that I add the data I used for the plots. Since that data was a little sparse, I currently have a run going to generate a more dense version, with 300 data points like the original plot.

Also on my to-do list: linear programming is the main thing that's been slowing down generating these bounds. If I switch over to using @uhrm's improved linear programming method, it should speed things up a lot.

evan-conway:
@teorth I've added the data under Data/t_large_n_bounds.txt, along with the following two graphs of $t(N)/N$ and $o(1)$:
t_large_n_bounds
o(1)_large_n_bounds

teorth:
@evan-conway and @matthieu-rosenfeld , many thanks for the data!

@matthieu-rosenfeld , please do submit the data under a separate PR.  I guess I misspoke when I asked for the number of factors greater than N/3; I meant the surplus number of factors exceeding N when applying the greedy algorithm to t=N/3, which I was hoping your algorithm would be able to compute readily, as a second measure of how far "above" N/3 one is.

As a sanity check, perhaps one could also run the greedy algorithm in the same range as @evan-conway 's linear programming data.  Presumably the greedy algorithm should fall a bit below the LP upper bound (if not, something has gone wrong), but this should be useful confirmation that everything is basically working as expected.

PR #24, opened 2025-04-16
Add A034258 values up to N=10,000 as computed by lp_solve.

plby:
Running lp_solve on the linear programs generated by the program from commit https://github.com/teorth/erdos-guy-selfridge/commit/7f837671ff858a4c4601f34b3af2e8cf09f73ae2 / PR https://github.com/teorth/erdos-guy-selfridge/pull/23 , I have computed the values of [A034258](https://oeis.org/A034258) up to $N=10\\,000$.  (This is the threshold function $t(N)$ we've been focusing on.)

So these are computed using lp_solve's integer programming ability.  They have not been confirmed with any kind of dual LP bound or anything, whether exact arithmetic or not.  My assessment is that the values are probably all exactly correct (especially for the purposes of OEIS), though depending on your mathematical standards, they might not be "proven".  I hope it's possible to extend this sequence up past $N=43\\,632$ in the same way, though it'll take some computation time.


evan-conway:
Nice work! Are you using the optimizations proposed by @uhrm? That might allow better scaling for large problems.

teorth:
Thanks!

Perhaps one can do some random spot testing of the upper bound to see how well it matches the lower bound data here.  I'd be cautious about adding a large unverified data set to the OEIS unless it is accompanied by a very visible disclaimer, and that it is not added directly to the primary OEIS listing but rather as a secondary link (with the disclaimer attached, namely that it is a verified lower bound but not a verified upper bound).

plby:
I should be more clear: these are exact values, ie matching lower and upper bounds. `lp_solve` solved all of these integer program problems to optimality.

The imperfect bits are: (1) we assume the factors are at most N, (2) `lp_solve` uses floating point arithmetic, and (3) the programs may have bugs. I think (1) is a real mathematical issue (which would be great to address!), but (2) and (3) are well within the acceptable limits for OEIS.

Because lp_solve prints a solution, the lower bound is certainly correct (and easy to verify, as I do in https://github.com/teorth/erdos-guy-selfridge/pull/5 ). When the (relaxed, ie non-integer) linear program upper bound is tight (ie within $1$), then lp_solve can produce a dual certificate that is also easy to verify, but there are small numerical concerns similar to those I address in https://github.com/teorth/erdos-guy-selfridge/pull/15 . This is the situation for the majority of the numbers computed here.

When the relaxed program isn't tight, the branch-and-bound algorithm (or other strategy) used by lp_solve still solves the problem optimally, ie "exhaustively". However, there is no longer a simple certificate to prove optimality: the lp_solve run *is* the certificate, modulo the issues (2) and (3).

plby:
All that said, I can collect data to indicate when the relaxed/non-integer linear program is tight.

teorth:
I see.  Since we do have a dual LP method that is not subject to (1) (namely, producing a dual certificate for factors up to N which is also monotone), perhaps there is some initial interval of N for which we can get completely validated exact values of t(N) (though given that N=155 had previously caused problems in this regard, perhaps the interval we get will be significantly shorter than N<10,000).    There is also a significantly slower finite guarantee if we allow for factors larger than N but still reduced in the sense of having no smaller factors greater than or equal to N, or no prime factors larger than N; this cuts down the number of factors to consider to at most N^2, which looks expensive but perhaps if there are only a small number of places where things are not tight then this might be ok.  In any case, we could have two data sets - a short, fully validated exact data set, and a longer data set which is fully validated as a lower bound and only heuristically validated as an upper bound.

plby:
I have been running the LP relaxation.  I have results up to $N\le 7\\,000$ now and expect to have $N\le 10\\,000$ by tomorrow.  Thusfar, all of the duals have been sorted.

As a preview of the results, there are only $10$ values of $N\le 7\\,000$ where the LP relaxation doesn't give a tight bound.  They are the following:
```
155 45 155.00000000 sorted
765 236 765.00909091 sorted
1528 481 1528.00000000 sorted
1618 510 1618.00000000 sorted
1619 510 1619.00000000 sorted
2574 821 2574.00000000 sorted
2935 938 2935.00000000 sorted
3265 1048 3265.00517241 sorted
5122 1653 5122.00335570 sorted
5680 1839 5680.00131926 sorted
```

The way to read this is that I believe $f(155)=45$, because that was the solution `lp_solve` found to the integer programming problems -- assuming all of the factors are at most $N$.  (And to be explicit: it provides a construction that achieves this.)  In particular, this means that according to `lp_solve`, there did not exist an integer solution to the LP for $N=155, T=46$ (note that $T$ is one higher here, because it's the first impossible one).  But there *was* a real (ie, non-integer) solution to the LP relaxation and it achieved an objective of exactly $155$.  This is exactly as @uhrm reports here: https://github.com/teorth/erdos-guy-selfridge/pull/1

(The `sorted` means that the dual was sorted the way we want it to be, which is to say that for the LP relaxation, it does not help to use factors greater than $N$.)

In short: the LP relaxation is *very* good in this range!

teorth:
Very nice!  It would great to somehow get definitive proof that $t(155)=45$ and $t(765)=236$, as then we could have a verified sequence of 1527 entries for the OEIS, which I think should be plenty.  The former bound $t(155)=45$ is already implicitly claimed in https://oeis.org/A034259 , and the integer programming tells us it is tight as long as we only use factors up to 155, but it would be good to confirm it without the restriction.  Doing this for 155 and 765 I think would be a satisfactory way to conclude Goal 3 of the project.

plby:
Thanks for merging this PR.  I made one more change over at https://github.com/teorth/erdos-guy-selfridge/pull/25 to improve the presentation of the LP bounds.  But there are some more notes there that describe how the OEIS sequence has now been verified up to $N\le 10\\,000$ except for $N=9\\,633$ (relevant to Goal 3 as you say).

PR #25, opened 2025-04-17
Add some lines to loose_lp_bounds.txt to include the actual LP relaxation bounds.

plby:
Okay, the LP bounds have been computed for all $N\le 10\\,000$.  There was one more value $N=9\\,633$ for which the LP bound is not tight.  I collected all $11$ of these in a file called `Data/loose_lp_bounds.txt` which made it into my latest commit in my previous pull request https://github.com/teorth/erdos-guy-selfridge/pull/24 which already got merged.  The format is very similar to what I wrote in my previous comment there: https://github.com/teorth/erdos-guy-selfridge/pull/24#issuecomment-2811388882
```
155 46 155.00000000 sorted
155 47 154.45454545 sorted
765 237 765.00909091 sorted
765 238 764.51818182 sorted
1528 482 1528.00000000 sorted
1528 483 1527.60000000 sorted
1618 511 1618.00000000 sorted
1618 512 1617.51111111 sorted
1619 511 1619.00000000 sorted
1619 512 1618.51111111 sorted
2574 822 2574.00000000 sorted
2574 823 2573.53448276 sorted
2935 939 2935.00000000 sorted
2935 940 2934.62184874 sorted
3265 1049 3265.00517241 sorted
3265 1050 3264.70517241 sorted
5122 1654 5122.00335570 sorted
5122 1655 5121.68120805 sorted
5680 1840 5680.00131926 sorted
5680 1842 5679.61345646 sorted
9633 3145 9633.00000000 sorted
9633 3146 9632.93968254 sorted
```

First of all, I've incremented $T$ on each line so that it's much more obvious to read now.  (That's already been committed, but only includes half of the lines above.)  Each line reports what the LP bound is for a specific $N$ and $T$, with no need to add $1$ in your head in any circumstance.  I think that is much more straightforward!  In this PR, I have also included the actual LP bound in each case.  For example, the second line reports that for $N=155, T=47$, the LP maximum is less than $N$ and thus that $(N,T)$ pair is infeasible even for real numbers.  Because the dual is sorted in all cases, this does not even assume the factors are all less than $N$.  Note that for $N=5680$, the LP bound is $2$ off from the integer lower bound as opposed to just $1$ in the other cases.

For all but $N=9\\,633$, I have successfully run the integer LP for the larger $O(n^2)$ size (primal) instance I describe here: https://terrytao.wordpress.com/2025/03/26/decomposing-a-factorial-into-large-factors/comment-page-1/#comment-687732

I have not yet understood the optimization @teorth  proposes in the response at https://terrytao.wordpress.com/2025/03/26/decomposing-a-factorial-into-large-factors/comment-page-1/#comment-687734 but I will do that next.

There have been no greater solutions for the big integer programs.  That is, for all $N\le 10\\,000$ (except maybe $N=9\\,633$ because it hasn't finished running), the following process has succeeded:
1. Compute the LP relaxation upper bound.  The program assumes all factors are at most $N$, but we can check whether the dual is sorted.  It has been in all cases, so this is an honest upper bound which might simply not be tight.  (For example, this proves $t(5\\,680)\le 1\\,841$.)
2. Compute the integer program lower bound, assuming all factors are at most $N$.  Because this is an explicit construction, this is an honest lower bound.  (For example, this proves $t(5\\,680)\ge 1\\,839$.)
3. If these match, great: this means the upper and lower bounds are proven tight!  Otherwise, solve the (much bigger) integer program *not* assuming all factors are at most $N$.  The result of step 3 has always matched the lower bound from step 2.  (For example, this proves $t(5\\,680)=1\\,839$.)

I will report when the big integer program for $N=9\\,633$ finishes as well.


plby:
My writing is a little unclear above.  For any given $(N,T)$ value, we get a value which represents the maximum to a particular (real) linear program; I am not referring to that as a *bound* above, even though of course it is both a lower bound (witnessed by the primal solution) and an upper bound (as witnessed by the dual solution) for the linear program.

What I mean by *bound* is that if this maximum is less than $N$, then this produces the bound $t(N)<T$ (either assuming the factors are less than $N$, or without that assumption in the case the dual is sorted).  I just wanted to leave this comment to make more clear what I mean by "actual LP relaxation bounds" in the title of this PR.

teorth:
OK, so just to confirm: the data given in https://github.com/teorth/erdos-guy-selfridge/blob/main/Data/a034258-lpsolve.txt is now guaranteed (by a combination of linear programs, dual linear programs, and integer programs) to be the exact value of t(N) for all N <= 10000, with the possible exception of 9633 where it is a lower bound (and exact if one restricts to factors at most N), but has not yet been confirmed as an unrestricted upper bound?

plby:
That is correct.  Also, the big integer program for 9633 finished, so that is verified now too.  That is, when asked to factor 9633 into as many terms as possible, each of which is at least 3145 (but are allowed to be bigger than 9633), it is only possible to find 9632 terms.  Amusingly, in the solution it chose with 9632 terms, lp_solve used 9840769 as one of the factors!  (This was the only factor used greater than 9633.)

A note: in the README right now there's a note that says $3145\le t(9633)\le 3146$, but actually the status before this verification was $3144\le t(9633)\le 3145$ and thus after verification it's now $3144=t(9633)$.  This is correctly reflected in the committed file https://github.com/teorth/erdos-guy-selfridge/blob/main/Data/a034258-lpsolve.txt#L9633 as well as above.

I assume I caused the confusion in part by my changing notation in this file `Data/loose_lp_bounds.txt`.  The meaning of these two lines
```
9633 3145 9633.00000000 sorted
9633 3146 9632.93968254 sorted
```
is that the LP relaxation *can* achieve 3145 (because $9633\ge N$) but *can't* achieve 3146 (because $9632.93968254<N$).  The integer program however can't get 3145.  The reason the first line is included is because the relaxation isn't tight, but we can see how much room there is: as little as possible.  The reason the second line is included is I wanted to show what the relaxation bound actually *was* since it isn't tight.  I assume following all the details in this note isn't particularly relevant because I presumably that comment in the README will be removed now that Goal 3 is in a very good place.

Tldr: $t(N)$ is computed and verified for all $N\le 10\\,000$.

plby:
A curiosity for $N>10\\,000$: I have now seen 9 LP duals that are not sorted.  The first one was $N=10844$ and $T=3552$ and the remaining ones were nearby.  I have not (yet) looked into them to see what's up; in particular, I haven't checked whether there *exists* a sorted dual, but it simply chose not one which isn't.

plby:
(The verification for 9633 did not use the idea from https://terrytao.wordpress.com/2025/03/26/decomposing-a-factorial-into-large-factors/comment-page-1/#comment-687734 though I do now understand it and think it would be quite helpful starting already from this value of $N$.)

teorth:
Great!  So I've updated the README to indicate that the N \leq 10000 data is now fully verified.  This seems like a good data set to submit to the OEIS, I will look into doing that now.

PR #28, opened 2025-04-19
Small Prime Optimizations

kevinventullo:
This adds some logic to improve the memory utilization and speed of the main algorithm.

The basic idea is that we are better off computing small prime factors of an integer on the fly rather than storing them. Naively, dropping multiples of p should reduce the memory footprint of the smallestDivisor hashmap by a factor of (p-1)/p. 

We can do this for multiple small prime factors at once by pre-computing an array equal to the size of their product. There are some memory/runtime tradeoffs here which are described in a bit more detail in the comments. 

I left a few comments for some additional optimizations as well, but this may be moot as @teorth seems to have gotten the theoretical bounds down to 10^11 anyway.

### DIFF
--- a/src/from_medium_to_large/RangeOfValues.cpp
+++ b/src/from_medium_to_large/RangeOfValues.cpp
@@ -23,25 +23,134 @@ cc_hash_table<long long, long long> largestDivisor;
 // this is the number of consecutive intervals that are used for precomputing primes and the factorization
 #define NUMBEROFINTERVALS 400
 
+
+// Chosen so that the product of the entries in a given list is not too large.
+// The current choice seems to work well. 
+//
+// Each list corresponds to a single array lookup for the primes covered.
+// So, e.g. checking divisibilty one prime at a time would roughly correspond to having
+// a separate list for each prime. For convenience, here is the % coverage for including all primes
+// up to p for the first several values of p:
+//  2 : 50.0 %
+//  3 : 66.7 %
+//  5 : 73.3 %
+//  7 : 77.1 %
+//  11 : 79.2 %
+//  13 : 80.8 %
+//  17 : 81.9 %
+//  19 : 82.9 %
+//  23 : 83.6 %
+//  29 : 84.2 %
+// 
+//  Note that getting above 90% requires the first 55 primes.
+// 
+// In general, more lists means more run-time lookups, but less memory used and hence better cacheability.
+// These can be rearranged or edited for tradeoffs. As written, the only logical requirement is they must 
+// cover exactly the first N primes for some N, and the primes must appear in order.
+const vector<vector<long long>> small_primes{
+    {2, 3, 5, 7}, {11, 13} 
+    // {17, 19},
+    // {23, 29, 31, 37},
+    // {41, 43, 47, 53}
+}; 
+
+// After running prepSmallPrimes, this will contain the same number of vectors as 
+// small_primes, but each one has length equal to the product of the primes in the corresponding
+// list in small_primes. The ith term of a sub-vector of small_prime_mods is
+// equal to the index in primes of the smallest prime that both divides i and belongs to the 
+// corresponding list in small_primes. Otherwise, -1 if i is not divisible by any of those primes. 
+// Thus, for any integer i > 0, if the smallest divisor of i is one of the small_primes,
+// the value can be determined by looking up, in order, the value of v[i%v.size()] for each v in small_prime_mods.
+vector<vector<long long>> small_prime_mods;
+
 vector<long long> primes;
 
+void prepSmallPrimes(){
+  // Populate directly as we will skip small primes in the main loop. 
+  for (const auto& v : small_primes) {
+    for (auto p : v) {
+      primes.push_back(p);
+    }
+  }
+  
+  long long curr_index = 0;
+
+  for (int i = 0; i < small_primes.size(); ++i){
+    long long vsize = 1;
+    const auto& curr_prime_list = small_primes[i];
+    // Compute the product of the primes in the sublist.
+    for (int j = 0; j < curr_prime_list.size(); ++j) {
+      vsize *= curr_prime_list[j];
+    }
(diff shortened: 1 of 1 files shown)


matthieu-rosenfeld:
Thanks for adding this optimization.

 I am running some "benchmark" to see how much it saves:
* for the interval $[10^7, 10^8]$: old code is 6 minutes, 1.4 GB, your code is 12 minutes 0.4GB
* for the interval $[10^7, 5\times 10^8]$: old code is 16 minutes, 2.8 GB, your code is 27 minutes 1.6GB
I am now running it on  $[10^7, 5\times 10^9]$ to have one more data point. It seems that the difference in memory usage now become really negligible (it is still running, but I would say 13GB against 12.5GB). The only explanation that I see would be that the $\frac{n\times P_+(n)}{P_-(n)}<n/3$ criterion already tends to eliminate most numbers that have a small divisor (and it is indeed reasonable to expect that it eliminates mostly numbers that have a small divisor). 

I think the easiest way to improve the storage of the factorizations is to find a better criterion to eliminate numbers. It could even be a heuristic criterion, in the sens that is eliminates factorizations that are "highly-likely" to be useless and then if we ever encounter such a number whose factorization has not been precomputed, we simply ignore it and go to the next number. 

The other way would be to try to go back to using a vector instead of a map to store the factorizations. In fact, since vector is much more efficient than a map "per value stored", it should almost fit in memory with this small prime optimization.For $10^{11}$, we need to store the factorization of only $10^{11} / (3\times 5)$ which should take "only" 50GB (and the ignore number divisible by 3,5,7,11 or 13 can be used with a vector or a collection of vectors). Also, if it is done right, I suppose it could be faster with a vector.

matthieu-rosenfeld:
So I was trying to understand why the small prime optimization helps less than expected for "large" $n$ by inspection of the factors that are actually used by the algorithm (I mean the "toBeRemoved" factors). I did something completely silly, which is to forbid the use of factors larger than $n/20$ if they are not divisible by $2$ ($20$ is an arbitrary value, but it is the one I experimented with). For large enough $n$, the greedy algorithm still finds a factorization (as expected the surplus is slightly smaller, but only slightly and we have a huge margin). This is the kind of criterion that we can use even if we store the factorization in a vector instead of a map ! As the value $20$ is completely arbitrary, we can probably replace it with something larger (we could probably even add numbers divisible by $3$ to increase $20$ even more).

So I changed the code to use a vector that store the factorization of odd numbers up to $n/20$. Also, since it is the index of one of the prime smaller than $n/20$, we can use a vector of int instead of a vector of long long. So for $n=10^{11}$, the factorization vector occupies $(10^{11}/40)*4B=10GB$ which is smaller than the $~32GB$ occupied by the prime numbers.  One can now easily run the code on a modern laptop for $10^{10}$ (I ran it for $10^9$ in 20 minutes over 1GB) . So now if we want to reach $10^{12}$ (I do not know if this is still necessary, I think the problem in the $10^{11}$ proof was fixed with interval arithmetic ?), the only issue would be storing the primes, but we could simply decide to store less prime and to only keep the prime counting function $\pi$ for large $n$ multiple of say $10000$.

I am currently running the code for $10^{11}$. I will make a PR with the new version of the code if no problem occurs.

teorth:
Yes, I believe the issues above 10^11 are now fixed with interval arithmetic, but it doesn't hurt to have some redundancy and safety margin, plus some of the optimization tricks that we discover here (and report upon in the paper) may be useful for future researchers working on similar problems, so it is still good to try to make the numerical methods faster and more lightweight (in particular running on a modern laptop in reasonable time would be a good target).

One thought I had was a hybrid greedy/linear programming approach in which one uses greedy methods to exhaust all the primes above a certain threshold, and then switches to a linear program for the few remaining primes (and some preselected combinations of those primes that are just barely above N/3) to close.  This doesn't help so much with the memory issues, and perhaps is a tiny bit slower than direct greedy, but might grab more of the available surplus that a direct greedy approach would concede, and might therefore permit one to cover ranges such as [10^11, 10^12] with fewer data points for N.  

If greedily clearing away the large primes is a computational bottleneck, one can also use explicit upper and lower bounds for primes in various intervals (which are now detailed in the current paper) to estimate their impact to high accuracy without actually having to count the primes directly.  For instance, to clear away the primes between t and N, one uses a lower bound for the number of such primes (see equation (2.16) of the paper) to just perform the "virtual greedy algorithm" of directly incrementing the counter for the number of terms by this conservative lower bound (and conceding any wasted surplus).  For the primes between t/2 and t, one does the same thing, while also decrementing the available powers of 2 by a conservative upper bound for the number of such primes.  (By the time one reaches 10^11, the gap between the upper and lower bounds is so slight in relative terms that I don't think much is lost by doing this).  One can keep doing this down to some threshold t/K for some reasonable sized k (E.g. k=100), at which point one can switch from the virtual greedy to an actual greedy algorithm that requires memory storage for the primes p up to t/K.

uhrm:
@teorth Regarding your idea of a hybrid greedy/linear programming approach: this is pretty much exactly what my `smoothfac` implementation does. It exhausts all primes larger than $`\sqrt{N}`$ greedily and then deals with the $`\sqrt{N}`$-smooth factors using a linear program. The sifting approach in my latest implementation is precisely a preselection of the smooth factors close to $t$ (and then iteratively adding more factors as long as the solution is not optimal).

The choice of $`\sqrt{N}`$ is empirically justified (and also helps with some technicalities in the implementation).

Btw. another observation that I made was that the optimal dual multipliers $`w_p`$ are very close to $\log{p}/\log{t}$ (with relative error $`\approx 10^{-4}`$. I wasn't able to use this fact, though.

teorth:
Ah, did not realize that, good to know!

Your observation that $w_p \approx \log p/\log t$ is very interesting.  I'll try to see if there is a theoretical explanation for it.  Possibly it could lead to an alternate proof of what is currently Lemma 5.3 in the notes.  In any case it helps explain why empirically $w_p$ is observed to be increasing.

PR #29, opened 2025-04-19
Grid search to find better parameters

evan-conway:
@teorth I did a basic grid search to try and find better parameters and was able to get things down to $N \geq 10^{10.7555}$ by setting $A = 186.5$, $K = 293$, $L = 4.5$. I'm not sure how much further this can be pushed, but hopefully this gives a little more room to work with. I've included the code for the search algorithm I was running.


teorth:
One caveat I should have mentioned: A and K are constrained to be integers, though L can be fractional.  Hopefully this does not cause a serious degradation in the bounds.

evan-conway:
While searching for a better value with $A$ as an integer, I stumbled upon a somewhat strange issue. For the values $A = 189, K = 293, L = 4.5$ and plugging in $N = 10^{10.69}$, the evaluate function does not trip any assert statements, indicating that it is a valid set of parameters for that $N$ and above. However, both $N = 10^{10.68}$ and $N = 10^{10.70}$ trip the assert statement regarding $\delta$. As a note, for each of these $N$ values, I cast it to an integer before passing it, so I do not anticipate that as the issue (the corresponding $N$ are $48977881936$, $47863009232$, and $50118723362$, respectively). This behavior seems to contradict the top of the document, which says that the bounds should hold for larger $N$, though I may be misinterpreting.

teorth:
Hmm, that is concerning, possibly a sign of a bug in my code.  I'll look into it, thanks!

teorth:
Oof, there is a real issue: delta_5 is not necessarily monotone (it is formed by taking the difference of two monotone quantities).  (And now I see that there is going to be a similar issue with delta_7.)  This makes things rather annoying - for large enough N I can afford to replace these non-monotone quantities by monotone ones (in fact my original arguments were basically of this type), but for the 10^11 type ranges that may be too expensive, and one may have to end up instead verifying intervals of N at a time with interval arithmetic, which is doable but significantly more tedious.

I'll have to think about fixes, but it may be that one has to retreat to a larger scale like 10^12 where there is more room and I can drop some of the negative terms that I had introduced in order to make things as small as possible (without realizing that this compromised monotonicity).  

Anyway, this is another example where numerical work helped uncover a theoretical issue, so even though it is at least temporary bad news, I'm grateful for your efforts in uncovering it!

teorth:
OK, I'm now seeing quite a bit of non-monotonicity in my bounds.  For any single N, the code should still be fine, but then it just verifies that one N and does not automatically guarantee the conjecture at larger values of N.  The non-monotonicity is not huge -it comes from the very slow variation of log(N) - but it is present, and so I have to downgrade my claim of having resolved the conjecture for all N >= 10^11 to having resolved it just at N = 10^11 with strong evidence of it being true for larger N as well.

I think I am going to have to bite the bullet and actually learn how to use the interval arithmetic library in Python, and then refactor the code to allow N (or log(N), really) to take values in an interval, and then all the deltas etc. will also come with their own interval.  It does at least have the silver lining of having eliminating any sources of false claims coming from roundoff error.  This is not extremely difficult, but it will be time consuming unfortunately...

teorth:
Actually, interval arithmetic isn't so bad!  I now have uploaded `interval_computations.py` which is a version of the previous `calculations.py` in which `N` is now an interval `mpi(N_1,N_2)` rather than a single number.  (The one wrinkle is that one has to work with the normalized version t_norm = t/N of t, which for the purposes of Conjecture 2 is simply 1/3.), and then the code verifies that t(N) >= t_norm N for all N in the interval [N_1,N_2].  

With the fixed choice of parameters A = 190, N = 252, L = 4.5, I can cover the range of N from 10^11 to 10^100 using five intervals: [10^11, 3 * 10^11], [3 * 10^11, 10^12], [10^12, 10^15], [10^15, 10^25], and [10^25, 10^100].  The verification gets increasingly easier as N gets larger, and the intervals that one can cover at a time become way larger.  Past 10^100 I should be able to find a simpler theoretical argument.  

I'm calling it a night for now, but tomorrow I'll try to see if I can simplify the argument here

teorth:
I've now uploaded a new version of the notes powered by the interval arithmetic calculations, which now actually verifies Conjecture 2 for all N >= 10^{11}, and not just the endpoint N = 10^{11}.

evan-conway:
Huh, that's interesting! I noticed when doing the initial optimization run that $K = 317$ would often pop up as well, so I'm guessing something similar happens there.

Also, I took a look at the interval arithmetic code, and with $A = 189, K = 293$, managed to get the lower bound down to $N = 6 \cdot 10^{10}$ while still covering the full upper region. This can possibly be pushed slightly further by playing around with different parameters that may be more favorable for the lower ranges, but I suspect you would have to automate the process in order to see meaningful gains, since increasingly small intervals are required in order to make progress.

teorth:
That's great news! I've updated the paper (and README) accordingly.  This gives us a bit of a safety margin with the numerical verifications coming from the greedy algorithm.

PR #30, opened 2025-04-21
Add results from LP runs with large N

uhrm:
I have worked on an implementation of the LP code that is able to scale pretty well. It is based on the ideas in the program `smoothfac.py`. I have noticed that when one sets the threshold of smoothness at $`\sqrt{N}`$ instead of the originally proposed $`\sqrt{t}`$, one is able to recover optimal solutions of the full LP (tested up to $`N = 9 \cdot 10^6`$ for which I could solve the full LP).

The code I've used to produce these results is implemented in [Mojo](https://docs.modular.com/mojo/manual/get-started/) and is optimized for storage. It also uses a custom sifting strategy for solving the LP to reduce the number of columns being added to the LP solver. (The implementation has become slightly nontrivial.)

In the file `lp_large_n.txt` the column `t(N)` is the largest $`t`$ for which a factorization with $`N`$ factors could be found (it is thus a lower bound). The column `ub` is an upper bound to the full LP (limited by the numerical accuracy of the dual multipliers returned by the LP solver, which appears to be ~10 decimal digits). I've also reported the asymptotic bound with constants $`c_0 = 0.304419`$, $`c_1=0.7555`$.

In principle, the code could be used to find good factorizations for larger $`N`$, but the estimates of the upper bound become unreliable.

I will also add the actual code to this repository. I've been developing it here: https://gitlab.com/uhrm/facfac-mojo.


evan-conway:
Agreed; having full data in that range would be quite useful! In particular, one thing I've been thinking about is the possibility of doing some kind of polynomial fit. If we are hypothesizing that the asymptotic form is $\frac{1}{e} - \frac{c_0}{\log(n)} - \frac{c_1}{\log(n)^2}  - \frac{c_2}{\log(n)^3} - ...$, then applying a $\frac{1}{\log(n)}$ transformation to the data and then fitting a polynomial to the results would be a good way to hypothesize values of $c_k$. This obviously requires care to not overfit, but should be able to at least get an estimate for $c_2$.

uhrm:
I forgot to mention that currently, the upper bounds my code computes are conservative. I collect them opportunistically while bisecting to optimize the lower bound. There might be a few more bisection steps necessary to also get the best upper bound. I would have to implement that, though.

evan-conway:
@teorth on the topic of $c_1$, have you uploaded the code used for your initial estimate from two days ago? It's fine if it's not available yet; I was just curious.

uhrm:
I've just pushed the results from my run for $`N \in [10^3, 10^5]`$ in steps of 100. I've also implemented proper bisecting for the upper bound.

The results seem consistent with the results in `t_large_n_bounds.txt` (by manually sampling some intervals). The only discrepancy I could find is the upper bound of the final value $`N = 10^5`$. I double-checked with the program `facfac.py` and the value 33646 seems to be a valid upper bound.

teorth:
> @teorth on the topic of c 1 , have you uploaded the code used for your initial estimate from two days ago? It's fine if it's not available yet; I was just curious.

My initial code is at https://github.com/teorth/erdos-guy-selfridge/blob/main/src/python/numerical%20integral.py but I am halfway through writing https://github.com/teorth/erdos-guy-selfridge/blob/main/src/python/interval_constants.py which redoes everything in interval arithmetic.  Right now I have a rigorous interval arithmetic computation for c_0; the same methods should also work for c_1, but I have to enter in the formulae, will do so later today when I have time.

teorth:
the interval_constants.py code is now complete for both c0 and c1 (to get adequate precision for c1 I needed to deploy the Erdos-Turan inequality to control error terms, but while the formulae were messy, they were just fine for computers to handle).  I now have eight significant figures on c_1: 0.75554808

teorth:
@uhrm I added your large N data as a table to the paper.  I also tried to compare with the upper bound that Lemma 5.1 of the paper gives (this lemma can be proven by linear programming, so it is worse than the best LP upper bound) but my implementation of the search was inefficient and I could only do the first 10 entries of the table this way.  Since it seems you already have some efficient code for bisection search and maybe precomputing primes, I was wondering if I could trouble you to complete this table?

uhrm:
Sure, I'll give it a go.

uhrm:
I got the following results:
```
      N       bound (5.1)

   100000        33668
   200000        67739
   300000       101945
   400000       136186
   500000       170502
   600000       204858
   700000       239241
   800000       273668
   900000       308099
  1000000       342567
  2000000       687883
  3000000      1034056
  4000000      1380747
  5000000      1727731
  6000000      2075114
  7000000      2422651
  8000000      2770389
  9000000      3118302
 10000000      3466414
 20000000      6952477
 30000000     10444694
 40000000     13940838
 50000000     17439638
 60000000     20940591
 70000000     24443233
 80000000     27947403
 90000000     31452859
100000000     34959207
200000000     70065426
300000000    105219155
400000000    140402099
500000000    175606238
600000000    210826906
700000000    246060998
800000000    281306449
900000000    316561839
```

teorth:
Thanks! I'll update the tables now.

teorth:
Hmm, there are two anomalies in the data now that I have built the table.  The linear programming bounds $t(8 \times 10^6) \leq 2771082$ and $t(4 \times 10^7) \leq 13942660$ are too large, in particular larger than the Lemma 5.1 bounds of 2770389 and 13940838 which (due to the new proof of this lemma via linear programming) is necessarily larger than the best linear programming upper bound.  All the other entries look to be in line, but something is wrong with these two.  Could I trouble you to run the linear programming upper bound again for these two values specifically?

uhrm:
Ah, yes. This run was done before I had the proper bisection for the upper bound implemented. I will re-run the whole range with the latest code. Then the anomalies should go away (barring any fundamental flaws in the implementation).

uhrm:
The results up to $`9 \times 10^7`$ are already in:
```
      N          t(N)          ub

   100000        33642        33646
   200000        67703        67704
   300000       101903       101906
   400000       136143       136149
   500000       170456       170459
   600000       204811       204815
   700000       239187       239196
   800000       273604       273610
   900000       308029       308042
  1000000       342505       342508
  2000000       687796       687800
  3000000      1033949      1033960
  4000000      1380625      1380637
  5000000      1727605      1727609
  6000000      2074962      2074983
  7000000      2422486      2422508
  8000000      2770212      2770241
  9000000      3118129      3118153
 10000000      3466235      3466247
 20000000      6952243      6952261
 30000000     10444441     10444454
 40000000     13940484     13940548
 50000000     17439282     17439315
 60000000     20940210     20940233
 70000000     24442818     24442855
 80000000     27946958     27947001
 90000000     31452431     31452454
```
This seems to resolve the anomalies.

The results for the $`10^8`$ range will take a good deal longer. I will update the file in the repository tomorrow when the series is complete.

PR #31, opened 2025-04-21
Contributions (uhrm)

uhrm:
I've added my contributions to the notes (and also fixed two typos).

Would you prefer to also have the compiled PDF included, or are edits to the source file ok?


teorth:
Thanks!  Either option is fine with me (I am continually recompiling the PDF anyways, so it doesn't matter much either way).

PR #32, opened 2025-04-22
3 versions of the C++ code

matthieu-rosenfeld:
While trying to understand why the small prime optimization did not work as well as expected for large values of $n$ (see PR #28 by @kevinventullo), I noticed that one could simply ignore the factorization of number $f$ larger than $n/200$ that are not divisible by $2$ or $3$. It leads to slightly smaller surplus, but it is negligible (well not exactly, but at least we still have a surplus, see one of my final comments about this). So we can now compute the factorization of way less integers. We can also only compute the factorization for odd number to decrease the size by another factor $2$ (we could even remove number divisible by $3$). Lastly, since we keep for each such $f<n/200$ the index of a prime that divides $f$, these $f$ are all small enough to fit in an int instead of a long long (which divides the memory consumption by $2$ again). Also, because of the factorization that we keep, we can go back to using a vector instead of a map for the factorizations. At least this is what I hopped.

The code RangeOfValuesNoMaps.cpp does this. On my laptop is uses $0.8GB$ during 4 minutes for the interval $[10^7, 10^9]$. This is much better than the previous version !! However, when I run it for $[10^7, 10^{11}]$, it seems to be much slower than the map version (by a few order of magnitudes). I tried a bunch of things, but was not able to figure out why exactly. My best guess is that maps are somehow much more cache-friendly than vectors for this kind of random access all over the place (??). I suspect this has to do with the different way arrays and maps are loaded to the cache (this is backed up by the fact that I tried to run it on two different big machines, and the more modern one with much bigger L2 and L3 caches seemed to be at least 50 times faster, but still really slow)

I can still add the same trick  (and remove the old criterion) to the map version RangeOfValues.cpp. It runs faster than the previous version, and it scales properly on the really large machine. It now takes 7 hours and 70GB (twice as fast and 3 times less memory used than the old version of the code). However, for the interval $[10^7, 10^9]$, it takes roughly twice as much time and memory use by RangeOfValuesNoMaps.cpp.

Based on @teorth comment on the related PR #28 we do not need to reach $10^{12}$. If we did the main issue would be to store the primes and the factorization of $n!$. But this could be easily resolved, by computing the primes only up to $10^{12}/9$ (or on an even smaller range), and using good approximations of the prime counting function $\pi$ to deal with the large primes. @teorth suggest we use 2.16 from the paper, I find 2.20 to be more useful (at least I know how to use it). I tried to implement a version of this in RangeOfValuesUsingPi.cpp. For the range, $[10^7, 10^9]$, it takes twice as much time but half the memory consumption than RangeOfValuesNoMaps.cpp. I think this version could scale nicely to $10^{12}$. I suspect that the main reason for which it is so much slower is the decrease of the surplus (see my comment below about the surplus).

I am not sure that we want to keep these 3 versions of the code, but I should at least share them here for those who might have suggestions or ideas (I did write a lot of other versions with all sorts of weird optimization that did not work over the last few days...).

Here are a few extra comments that concern the 3 versions of the code:
* Another optimization suggested by @evan-conway in PR #14 is to use much smaller windows in the segmented sieve to improve cache locality in the precomputation. It seems to make a huge difference, as the first phase is now also much faster.

* [The surplus:] this optimization (and the approximation of pi to handle big primes even more), decreases the surplus. This has two consequences on the running time. The first obvious one is that if we have less surplus, the next integer for which we need to factorize $n!$ is going to be smaller, so we need to factorize more. The second less obvious consequence is that we fail more when we try to factorize (since we start with a large expected surplus, and we decrease it until we succeed). One might expect that if we have to run it three times the cost is 3 times higher, but this is much slower when we fail than when we succeed (I think it has to do with the fact that when we succeed before the end, and the end of the greedy procedure is much slower). So I have tried to do some tuning of the choice of eps, to decrease this cost (it should help more to do it for the version that uses approximation of pi, since we lose even more surplus).

* Both the values of the segment size (NUMBEROFINTERVALS 10000) and the cutoff for factorization (LARGEST_COMPUTED_FACTORIZATION maxN/200) are chosen experimentally/arbitrarily. Both of these should be a function of maxN, and I did not try to really optimize these even for the current value $10^{11}$, but I do not expect that we can save much by optimizing it.

* Even if for some reason the small prime optimization does not help much, I included some of the other improvements by @kevinventullo in PR #28


teorth:
Thanks for all the code and experiments!  I now see there is a tradeoff between optimizing the speed of each individual run, and minimizing the total number of runs to be made, so not every trick to improve the former is a net win.

The bound (2.20) is not as accurate as (2.16) and so may cost too much surplus.  Because of the convexity of 1/log t, it is possible to replace the 1/log x factor in (2.20) with 1/log((x+y)/2) in the main term (at the mild cost of replacing log x with log y in the error term, which is much smaller in practice), but perhaps this only has a negligible impact on performance.  One could also lower bound the logarithmic integral in (2.16) with a simple Riemann sum; this basically amounts to subdividing [y,x] into some small number of intervals, applying (2.20) (or the above slight refinement) on each subinterval, and then summing.  If the loss of surplus coming from (2.16) is a bottleneck then this may help.

matthieu-rosenfeld:
I replaced the 2.20 bound by $(1-2/\sqrt{x})\frac{x-y}{\log((x+y)/2)} - 2\frac{E(x)}{\log y}$ (just to double check that it is what you suggested. It does make a huge difference, changing nothing else from the code it went from 25min to 8min40s for the interval $[10^7,10^9]$ (the version without this takes 7min40s, but uses twice as much memory). I will try to run this version for $[10^7,10^{11}]$ over the night.

I wanted to have a feeling for how good is this approximation. For $y= 3\times 10^9$ and $x=4\times 10^9$ (that could be typically the values we use it on), $\pi(x)-\pi(y)=4.5512\times10^7$, (2.20) gives $4.52\times10^7$, improved (2.20) gives $4.5497\times10^7$ and (2.16) gives $4.550\times10^7$ (I have similar results for a bunch of other ""typical" values). So replacing the improved version of (2.20), by (2.16) would probably require being quite precise to obtain any improvement.

By the way, the current implementation of this code, uses floating point arithmetic for $\sqrt{\cdot}$ and $\log$, so there is a small issue of correctness here. There might also be some -1 issues on where the limits of some intervals are as I was a bit careless, but nothing that a careful proofreading can't fix.

teorth:
The sqrt(x) term should be sqrt(y), but this should only have a negligible impact on performance.

The performance improvement is better than expected!  I am now tempted to go back to my numerical verification of the N >= 10^11 range, which is also basically using (2.19) and (2.20), and replace them with the sharper versions.  If nothing else it should give a bit more of a safety margin to guard against various errors (I'm now using 50 digit interval arithmetic for the N >= 10^11 stuff, so I'm pretty confident here that roundoff is no longer an issue, but of course there could be typos.)

evan-conway:
@matthieu-rosenfeld I forgot to mention this previously, but one thing that you might want to consider is splitting up the factorization of the initial range as well as the later range. To give a practical example, of this, the code currently has NUMBEROFINTERVALS set to $10^4$. If you were looking at the range $[10^7, 10^9]$, then this means that you would be computing one large block of size $10^7$, followed by $10^4$ blocks of size $\frac{10^9 - 10^7}{10^4} \approx 10^5$.

This is particularly important since currently the first block is greater than what fits in the L1 data cache, which I believe causes a significant slowdown. As a lower bound, a block of $10^7$ integers would be around $4 \cdot 10^7 \approx 39062.5 \text{ KB}$, which is well over what would be expected to fit in the L1 data cache.

matthieu-rosenfeld:
One more code ([RangeOfValuesNoMapsUsingPi.cpp](https://github.com/teorth/erdos-guy-selfridge/pull/32/commits/9f97c6c1929179cfcb49efbef344d6b2eacea33c#diff-c717172d32e78e9a2bf7cbfbaaf4577fbe737f7875c72e8c797591b509de6aa2)): I use vectors instead of maps to store the factorization, AND I use the approximation of the prime counting function. This version of the code now runs on my laptop in 30 minutes using 4GB of RAM for the interval $[10^7, 10^{11}]$. 

I have also joined the output of the code (although, anybody with a recent laptop and a C++ compiler can run it). It prints the remaining factors of $n!/B$ after each factorization. For instance, for the last factorization it prints:
>Donne for n = 98260542069 with eps = 30 next n = 101208358332
>Each of the factors found where larger than: 33736119443 (n/3 = 32753514023).
>Remaining factors: [5:465529099, 7:14443110, 23:29511539, 59:51386120, 61:89465830, 79:27308281, 101:2304802, 149:8257483, 163:41371107, 191:53158893, 257:6552527, 281:14897403, 331:36710378, 337:7929287]

The last line means that $n!/B$ is divisible by $5^{465529099}$ and the other numbers with their respective powers, which is a lot. However, it is not so surprising as we are asking for a multiplicative surplus of $1.03$ and we know that a surplus of $1.06$ is possible (by the previous version of the code).  So we could expect $n!/B$ to be as large as $(1.06/1.03)^{10^{11}}> 10^{10^9}$ which is much larger than the printed surplus. 

___

Most of the new implementation details are already mentioned in the first comment. It now  also uses the better approximation for $\pi(x)-\pi(y)$ mentionned in the previous comments. However, there is one more important detail:  decrease the value of the surplus that we are trying to reach. Replacing the target of $(1.06/3)$ by $(1.03/3)$, means we have to do almost twice as many factorizations, but these factorizations are much faster (and this is true with every version of the program). As already noticed by others, the end of the greedy algorithm is the slowest part, but if we have a lot of margin we do not need to reach the end of the algorithm (also for the same reason failure is really slow went it happens). I could be event faster if we had a list of precomputed surplus that we know the algorithm can reach for different values, but I do not think it is worth it.

___

I suggest that for now we keep this PR open, in case other ideas to improve the code are suggested, and I could then make a PR to add only the better version of the code. Note that this version of the code does not work for the interval $[80000,10^6]$, as I used the "weird heuristics" for this. I can at some point also provide a code for this range, if we still need it.


@evan-conway I added the suggested improvement, but it does not make a real difference as the first part of the computation is a really tiny portion of the computation time. But it makes the code more "uniform" so I guess it is better anyway.

evan-conway:
@matthieu-rosenfeld Yeah, realized that the speedup will be relatively negligible for cases like $[10^7, 10^9]$. In that case, only around 1% of the range is in the first block, so speeding it up wouldn't make a big difference. The impact will presumably grow larger for cases where you're examining a smaller range, since in that case the first block will make up a proportionally larger component of the total range.

PR #33, opened 2025-04-23
Update LP upper bound data of large-N run

uhrm:
This PR adds the new LP upper bound data obtained by the implementation that also optimizes the upper bound calculation.


teorth:
Thanks!  I'm still amazed at how accurate the LP method is asymptotically.

PR #35, opened 2025-04-26
fastegs

AndrewVSutherland:
This PR adds a fast, memory efficient C implementation of a modified version of the greedy algorithm that can verify the Erdos-Guy-Selfridge conjecture in the range [10^6,10^12] in about a minute using about 500MB RAM; this exploits a list of precomputed hints (optimal t-values to use for various N spanning this interval).  The time to compute the hints is about an hour (with the same memory footprint), and about five minutes for the interval [10^6,10^11].

I realize we no longer need to go up to 10^12, but I figured it wouldn't hurt to have some margin for error, and I think it's useful to have code that can quickly verify the range we do need using a small amount of memory.

I should note that this code is optimized to handle large values of N quickly, the lower bounds it gives on t are not as good as the standard, not maximizing t, so it won't do as well in the lower ranges (e.g. it only gives 101 extra factors for N=300000 and t=100000) and it cannot be used to verify the entire range [300000,1000000].


teorth:
Thanks, this is great! I've added you as a coauthor and collaborator.

One useful data set would be to report what lower bound on t(N) this method can give for round number N (of the form $a \times 10^j$ for $6 \leq j \leq 11$ and $1 \leq a \leq 9$), so that we can extend (a) Table 1 of the paper to compare this against the linear programming data, and (b) obtain some log-scale plot of t(N)/N in the range $10^6 \leq N \leq 10^{12}$ where currently we have only a small amount of data.   I guess this would require running a modified program in which one has to do bisection search or something to find the optimal t(N) rather than rely entirely on precomputed hints.

The other thing is to add to Section 3 of the paper (at notes.tex) to describe this alternate greedy method and its tradeoffs (faster, but not as optimal in t(N), and only competitive for large N).

AndrewVSutherland:
Will do, and there is an "or something" that is slightly better than a bisection search that the code uses to compute the hints.  It starts with $$t = \lceil\frac{N}{3}\rceil$$ and then based on the number of excess/missing factors, picks a new value of t. In combination with some smoothing to make sure it doesn't get stuck oscillating around the best bound, it converges quite quickly (substantially less than $$log_2(N)$$ iterations in most cases).

But this won't necessarily give the best possible lower bound on $$t(N)$$ the algorithm could potentially produce, because the bounds produced by this algorithm (and also the original greedy algorithm) are not monotonic in $$t$$ (for fixed $$N$$).

I don't actually know an efficient algorithm (say one that tests only a poly-logarithmic number of values of $$t$$) to compute the maximal $$t$$ for which the the algorithm succeeds on inputs $$N$$ and $$t$$.

One can compute an upper bound by looking at the number of factors the algorithm has found up to the point where it is always able to use the optimal m and then assuming you can perfectly split the remaining part of $$N!$$.  For  fixed $$N$$ this will will give an upper bound on the number of factors the algorithm could possibly produce for any given $$t$$ that is monotonically decreasing as a function of $$t$$, allowing one to do a bisection search on this upper bound.  One would then need to check ever $$t$$ in the gap between the value given by the computation of the hint and this upper bound.  I'm not sure how feasible this is, but I will give it a try.

I'm also happy to add a section to Section 3 explaining this algorithm and the tradeoffs in efficiency/memory (the key point is that if your goal is to check a large interval $$[N_0,N_1]$$ as quickly as possible, it's better to use slightly more $$(N,t)$$ pairs that you can compute quickly, rather than trying to get the best $$t$$ for each $$N$$, as I think has already been observed by others).

teorth:
I guess it's fine to report non-provably optimal lower bounds on t(N) from this approach (e.g., from the heuristic hint method and some backtracking) so long as the methodology is reproducible (and not excessively ad hoc).  I expect that while the method is technically non-monotone in t, assuming incorrectly that it is monotone should only cause a negligible impact on the t(N) bounds in practice.

Certainly a good idea to talk about tradeoffs!  These could all potentially be useful guidance for any future researchers who for some reason have to work with a knapsack type problem in which there are large numbers of small items together with a few very large items.

AndrewVSutherland:
Okay, I've change the code to let you compute bounds on $$t(N)$$ using the same algorithm that is used to compute hint-files, and I've changed the CLI so that it also supports a `-o` option that will ask it to compute an optimal bound (meaning optimal for the algorithm, not the actual value of $$t(N)$$), using the approach I suggested above.  As I feared, it isn't really feasible for large $$N$$, but running it on moderate values of $$N$$ confirms what you suggested above, the difference is pretty small.  On the other hand it allows the algorithm toi cover the range $$[530000,10^{6}]$$ whereas it would previously get stuck at several points in that interval.

I added a file with bounds on $$t(c10^n)$$ for $$1\le c\le 9$$ and $$6\le n\le 11$$, and also a file with optimal bounds for $$6\le n \le 8$$, that I can use to create a table.  Here are a few values for comparison (last column contains values computing using the `-o` flag):

| $$N$$ | $$t(N)\ge $$ | $$t(N)\ge $$ |
|---|---|---|
| $$3\times 10^5$$ | $$100408$$ | $$100408$$
| $$10^6$$ | $$337642$$ | $$338782$$|
| $$10^7$$ | $$3427641$$ | $$3433084$$ |
| $$10^8$$ | $$34621266$$ | $$34666162$$ |

teorth:
Thanks!  For N=10^5 the answer for tbounds_o is extremely close to N/3, is this an artefact of the code or does it reflect the true limit of the algorithm?

AndrewVSutherland:
Actually the value for N=10^5 was incorrect, because the code wasn't correctly handling the situation where it cannot prove t(N) >= N/3.  I have fixed this so that it will now handle any t >= N/4 for all N >= 45 (up to 2^40) and updated the value for N=10^5 in tbounds_o.txt, and added values for N=10^5 and 2*10^5 to tbounds.txt.  Note that the optimal bounds will  match the heuristic bounds in any situation where the optimal bound is < ceil(N/3) because the code computes an initial lower bound by starting with ceil(N/3) and working down from there until it works.

I also noticed a minor bug that in the edge case where the algorithm produces exactly N factors for a particular value of t was being treated as a failure to prove t(N) >= t rather than a success.  I've fixed this and am in the process of recomputing  the files tbounds.txt and tbounds_o.txt (the bounds in tbounds.txt have all been updated).

I'm also going to recompute the hint files so they match the new code.

AndrewVSutherland:
I've updated the one hint file that changed (3e5 to 1e6 where I used the optimized bounds to get the hints), the others were not effected by the < N vs <= N test.

I think this PR is ready to be merged.

Issue #36, opened 2025-04-28
Request for numerical data near N = 43632


teorth:
Hi everyone,

I was looking at the various data available on t(N) that we can use to display in the paper and I thought one nice addition would be some visualization of how the t(N) >= N/3 threshold is crossed near N = 43632.  In particular, I wonder if it is possible to obtain, for all N in a range near 43632, e.g., N in [40000, 45000], (which would also cover the first time N=41006 that Conjecture 2 holds), what the linear programming bounds are for the longest factorization of N! where all numbers are at least N/3, and (if computationally feasible) what the integer programming bounds here are as well (if this is not feasible for all N in this range, then maybe just a sampled range of N, e.g., the multiples of 10, would suffice).

Another related statistic in this range that one could gather is the upper and lower bounds on t(N) coming from the linear programming method, though again if this is too computationally expensive to do for all N, a sampled range of N would suffice.  (This would basically be a zoomed in version of what is currently Figure 6 at the point where t(N)/N crosses the 1/3 threshold.)

evan-conway:
Hmm, I'll give this a shot using the existing linear programming tools. If I'm understanding the requirements correctly, then I think I can get the desired bounds on longest factorization in a reasonable amount of time, though integer programming may take quite a bit longer. I suspect that the zoomed in version of Figure 6 should also be fairly reasonable, using @uhrm's Mojo code.

PR #37, opened 2025-04-28
Examining the [40000, 45000] Region

evan-conway:
In #36, @teorth requested some information about $N \in [40000, 45000]$. I modified smoothfac to return information about the lower and upper bounds on the length of the longest factorization with all factors $\geq N/3$, and wrote some code to report those bounds for a particular region. The bounds for $N \in [40000, 45000]$ are in the file crossover_bounds.txt under the data folder. This ran quite quickly, thanks to @uhrm's efficient smoothfac code.

The bounds are actually extremely tight for this region, normally identifying exactly the length of the longest factorization. If the exact values are required, it would probably be pretty simple to find all the cases where the length of the longest factorization is not exactly identified, and then run integer programming on those values.


evan-conway:
I graphed the difference between the upper bound on the factorization length and $N$, which produces this interesting-looking plot:
image

teorth:
Thanks for the quick work!

teorth:
I'm looking at the data now and there is something weird about the lower bound - there are many N > 43632 for which the number of factors is strictly less than N, e.g., at N=44424 only 44423 factors were reported.  Here is the plot of the lower bound as well as the upper bound:
Figure_1
But we did use the LP method to verify $t(N) \geq N/3$ for $43632 \leq N \leq 80000$ so I wonder if there is some discrepancy in how the LP code was used...

teorth:
OK, I *think* the issue is that sometimes, the lower bound for $t = \lceil N/3 \rceil$ doesn't reach N factors, but weirdly, setting $t$ to be a bit larger, say $t = \lceil N/3\rceil+1$ or $t = \lceil N/3\rceil + 2$, can work; the upper bounds should be monotone in the sense that raising $t$ can't increase the number of factors, but in theory the lower bounds could be non-monotone.  It's possible that the earlier LP runs by @plby and @uhrm also tested other values of $t$ than $t = \lceil N/3 \rceil$ which is why they could succeed for some $N$ that didn't work here.  (Just to pick one concrete example: the current run fails for $N = 44065$ with $t = \lceil N/3 \rceil = 14689$, but in https://github.com/teorth/erdos-guy-selfridge/blob/main/Data/factorizations/44065-44065-14690.txt.gz a factorization is given for this value of $N$ but with $t = \lceil N/3 \rceil+1 = 14690$.)

evan-conway:
@teorth I'm not sure if that's actually a problem. To the best of my knowledge, the way that Markus's smoothfac code works is:
1. Factor out non-smooth factors
2. Run a linear program for factoring smooth factors: this gives an integer lower bound (best factorization found) in addition to a real upper bound.
3. Take the best factorization found and run it through a greedy procedure. This improves the lower bound.

What I'm reporting in the data is:
- Lower bound = (# non-smooth factors) + (# factors in LP best factorization) + (# extra factors found by greedy run)
- Upper bound = (# non-smooth factors) + (# LP upper bound on remainder)

This means that the lower bound not reaching $N$ is just a consequence of the greedy algorithm not finding every factor it could, and the lower bound not being terribly good. This is why I just reported the upper bound, since my intuition is that it's more faithful to the actual length of the longest factorization in this range.

plby:
I do think there have been multiple different "lower bounds" from different methods, and they've been conflated a bit.  I describe one issue in https://github.com/teorth/erdos-guy-selfridge/issues/39 , but I think some of the uses are:
1. The lower bound method described in the paper, which is to take the optimal real solution and take floors.
2. The lower bound obtained by `smoothfac`, which is described above.  In particular, this uses a different linear program than item 1.
3. A lower bound obtained by a greedy method or modifications.
4. The lower bound obtained by attempting to solve an integer program corresponding to item 1.

The good news is these are all in fact lower bounds.  One concern I have is with the upper bounds.  The upper bound reported by the smoothfac method isn't actually an upper bound, right?

teorth:
Ah, this begins to clarify things.  It might be interesting then to run the original LP upper and lower bounds on the same range for comparison.  Based on the discussion, I'm imagining the ordering to be

`smoothfac` lower bound <= vanilla LP lower bound <= `smoothfac` "upper bound" <= vanilla LP upper bound

though as Boris points out the `smoothfac` upper bound is technically not an upper bound for the original problem, but for the problem restricted to factorizations in which the non-smooth numbers are extracted greedily.  I assume the vanilla LP lower bound should give t(N) >= N/3 for all 43632 <= N <= 45000 even if the `smoothfac` lower bound fails to do so.

AndrewVSutherland:
The bounds produced by egs -o N for N in 43523..45000 are short of N/3 by somewhere between 135 and 231 (they fluctuate within this range).

It occurs to me that for moderate size N (certainly up to 10^7, possibly 10^8) it would not be hard for me to extend the fastegs code to also implement the basic greedy algorithm with a -o option (telling you the best bound on t(n) the greedy algorithm can produce for this N).  I can give that a try tomorrow.

PR #38, opened 2025-04-28
Add lemma on weakly increasing dual weights

uhrm:
*Disclaimer:* Take this PR with a big grain of salt since I'm not a professional mathematician.

I think I found an argument to show that the optimal dual weights of the linear program are weakly increasing. It makes use of the fact that linear programs have basic solutions at which some inequality constraints hold with equality. It uses this to derive a contradiction that when the optimal weights are not weakly increasing they are infeasible (and thus not optimal).

There is currently the problem with the argument that I'm stating the final inequality without proof. It is easy to verify numerically, but I'm so far unable to find a mathematical formulation for its truth.

I though I publish this PR anyways, since if the argument holds, it could be interesting to add to the main text. It would put the numerical calculations on a more solid foundation.


teorth:
Thanks for the argument.  I have some questions:

1.  When optimizing linear program (4.2), (4.3), is j ranging over J_{t,N}, or the numbers between t and N, or all j?  You replace j with jq/p during the argument, which is fine if you are allowing j to be unrestricted, but in the other cases it is not necessarily the case that jq/p is going to be one of the j's considered in the linear program.
2. It should be possible to show your inequality $q^{\nu_q(N!)} \geq N/2$ for any $q \leq N$, once $N$ gets large (this is true for $q \geq N/2$, and for $q<N/2$ one has at least two factors so this covers the $q \geq \sqrt{N}$ case, and for $q < \sqrt{N}$ one should have at least $\sqrt{N}$ factors which should be more than enough since $2^{\sqrt{N}} > N/2$ for any reasonably sized $N$).  But I'm not sure how this rules out that $j$ is a basic factor, could you elaborate on that?

uhrm:
Addressing your questions:
1. Yes, the idea is to consider all $`t \leq j \leq N!`$ that divide $`N!`$.
2. The argument is essentially the same that you make to justify the domain $`J_{t,N}`$. The factor $`j`$ by assumption contains a factor $`p`$ and a factor $`q^{\nu_q(N!)} > N/2 \geq t`$. The factor $`p`$ can be dropped and $`j/p`$ is still a constraint of the LP. This implies that constraint $`j`$ cannot possibly be binding.

Thanks for the help with the inequality.

teorth:
Got it, thanks.  I'll merge and then edit your proof to clarify these points.

PR #40, opened 2025-04-29
Compute LP bounds from 40000 to 45000.

plby:
This is in response to the request https://github.com/teorth/erdos-guy-selfridge/issues/36 and the discussion in https://github.com/teorth/erdos-guy-selfridge/pull/37

I've computed the LP lower and upper bounds for $N\in\[40000,45000\]$ and $t=\lceil N/3\rceil$.  The file looks like this:
```
43631 14544 43624 43630.96260939 sorted
43632 14544 43626 43632.07716786 sorted
```
The columns are $N$, $t$, the LP lower bound, the LP upper bound, and whether or not the dual weights in the LP upper bound are sorted.  (They are all sorted in this range.)  The LP lower bound here *is* what is described in the paper, ie taking the floors of an optimal real solution.  In particular, what we see from the two lines above is that the LP upper bound does prove that $t(N)<N/3$ for $N=43631$.  However, this LP lower bound does not prove that $t(n)\ge N/3$ for $N=43632$, even though it is in fact the case; this is because this LP lower bound is not precise enough.


plby:
Sorry, I closed this because of a combination of git issues and forgetting to include 13 lines in the middle of the file!

Issue #39, opened 2025-04-29
Linear programming lower bounds


plby:
The paper currently describes the lower bounds obtained by the linear programming method as taking the optimal real solution $m_j$ and then taking its floors $\lfloor m_j\rfloor$.  If the sum of the $\floor m_j\rfloor$ is sufficiently large, then this is an admissible solution to whatever problem is being solved.  I guess this is what was done sometimes.

But not always.  A lot of the lower bounds I reported came from the solutions lp_solve gave when I asked it to solve an *integer* programming maximization problem.  These solutions are all admissible, and thus verifiably correct (I verified the ones in `Data/factorizations` with a simple program), but they don't *necessarily* match the description currently in the paper.  To be extra explicit: there is no correctness issue here.

I will do two things:
(a) see what the lower bounds actually are if you just take floors (most should be the same, but maybe they all are?), and
(b) adjust the statements in the paper if necessary.

teorth:
Thanks for looking into this.  One can certainly clarify the paper to say that there are at least two ways to use LP/IP to get lower bounds: the "cheap" way of taking integer parts of the real optimizer, and the "expensive" way of running a full integer program.  If there is some range of N where the two can be compared with an interesting graph, that would be nice (one doesn't necessarily have to do this for all the graphs in the paper, but maybe just one of them).

uhrm:
There is some  middle ground between the "cheap" and "expensive" ways: After rounding down the LP solution, one ends up with some remaining divisors that one can heuristically assemble to find further factors above the threshold. All the lower bounds that I reported do this (I describe it as 'Phase 3' a little bit [here](https://github.com/teorth/erdos-guy-selfridge/blob/main/src/mojo/smoothfac.pdf)).

For the large $`N`$ LP runs, using the "cheap" solutions directly would (roughly) double the gap to the upper bound (which would still be pretty good). For the smaller $`N`$ it is actually quite crucial to do this to get tight bounds -- the "cheap" LP lower bound would never be tight.

teorth:
I see.  This would be great additions to the LP sections of the paper.  Perhaps we could test all of these different LP variants on the same range $N = [40000, 45000]$, $t = \lceil N/3 \rceil$ (sampling $N$ to say multiples of 10 if necessary) to compare their performance.

plby:
[...]

PR #41, opened 2025-04-29
Compute LP bounds from 40000 to 45000.

plby:
This is in response to the request https://github.com/teorth/erdos-guy-selfridge/issues/36 and the discussion in https://github.com/teorth/erdos-guy-selfridge/pull/37

I've computed the LP lower and upper bounds for $N \in \[ 40000 , 45000 \]$ and $t = \lceil N / 3 \rceil$. The file looks like this:
```
43631 14544 43624 43630.96260939 sorted
43632 14544 43626 43632.07716786 sorted
```
The columns are $N$, $t$, the LP lower bound, the LP upper bound, and whether or not the dual weights in the LP upper bound are sorted. (They are all sorted in this range.) The LP lower bound here is what is described in the paper, ie taking the floors of an optimal real solution. In particular, what we see from the two lines above is that the LP upper bound does prove that $t ( N ) < N / 3$ for $N = 43631$. However, this LP lower bound does not prove that $t ( n ) \ge N / 3$ for $N = 43632$, even though it is in fact the case; this is because this LP lower bound is not precise enough.


teorth:
Interesting! in fact it seems that the cheap "floor" method just barely fails to demonstrate $t(N) \geq N/3$ for *any* $N$ in this range $[40000, 45000]$, though it comes very close to doing so at the right endpoint.  So what is the method that we do use to verify $t(N) \geq N/3$ in the range $43632 \leq N \leq 45000$?  I know we do have explicit factorizations in the data folder for this range, but now I don't fully understand how they were generated.

plby:
I asked lp_solve to solve the same maximization problem, but assuming the variables are integers.  (This is available as an option `-i` to https://github.com/teorth/erdos-guy-selfridge/blob/main/src/linprog/generate_lp_format.cc )

In other words, instead of naively computing $\lceil m_j\rceil$ because that's a feasible integer solution, lp_solve actually finds the integer optimum.

@uhrm notes at https://github.com/teorth/erdos-guy-selfridge/issues/39#issuecomment-2837779779 that the flooring method does not give tight bounds for $t(n)$ using the LP method even for small $n$.  (I will confirm this shortly.)  An alternative is described there (heuristically adding back factors).

plby:
[I mean $\lfloor m_j\rfloor$ of course.]

teorth:
Some basic plots: here is the vanilla LP upper bound and floor lower bound.  So the lower bound is somewhat worse than the smoothfac lower bound.
Figure_2
On the other hand, the LP upper bound and smoothfac upper bound are virtually identical:
Figure_1
Is it possible to also get the integer programming bounds in this range?  I can infer them for $N43632 \leq N \leq 45000$ from the file names in https://github.com/teorth/erdos-guy-selfridge/tree/main/Data/factorizations, at least when N = 1 mod 3, but this doesn't cover the full range.

plby:
Yes, I plan to compute the integer programming solutions right now.  To be clear, I mean the integer programming solution to the problem being discussed here: the maximum number of $N/3$-admissible factors of $N!$.

(I might actually compute $t(N)$ in this range as well, though it is significantly more computation.)

PR #42, opened 2025-04-29
Compute LP bounds for (N/3)-admissible factorizations from 4 to 45000.

plby:
Increased the range from [40000,45000] to [4,45000].  (The bound doesn't make sense for $N\le 3$ because $\lceil N/3\rceil=1$ which allows for an unbounded number of admissible factors.)


plby:
Okay, I've added the LP floor, IP lower, and LP upper bounds for all N from 4 to 55000 in the latest commit above.

plby:
The latest commit includes data up to 60000.  The IP lower bounds are getting worse, presumably because I'm hitting my computation time limits too often now.

PR #43, opened 2025-04-29
Add some integer programming bounds for (N/3)-admissible stuff.

plby:
I computed some integer programming optima for $(N/3)$-admissible factorizations for the range $[40000,45000]$.

However, I only tried each $N$ for one minute at most.  This succeeded for just over half of the numbers, and in all of these cases, the integer program achieved the floor of the LP upper bound.  The file now looks like this:
```
43630 14544 43625 ? 43629.84884646 sorted
43631 14544 43624 43630 43630.96260939 sorted
43632 14544 43626 43632 43632.07716786 sorted
43633 14545 43629 43633 43633.07716786 sorted
```
The columns are $N$, $T$, the LP lower bound, the IP optimum, the LP upper bound, and the sortedness of the dual.  The IP optimum is `?` if the run took over a minute.


plby:
For the factorizations in https://github.com/teorth/erdos-guy-selfridge/tree/main/Data/factorizations , I used the following process: for a given $N$, try to solve the integer programming problem for as high a $T$ as possible.  If an admissible solution is not found (either because one doesn't exist or it took too long), decrease $T$ and try again.  If a solution is found, then you can jump up to $N=3T+1$.  (In particular this means that all $N$ after the first are $1\bmod 3$, but they also skip more than that.)

Most of the time, the optimal $T$ was found, but sometimes it took too long and it was necessary to back off.  (I think $t(43885) \ge 14633$ is the first case of a potentially sub-optimal bound.)  I think a couple times I also manually decreased $N$.

So the data in https://github.com/teorth/erdos-guy-selfridge/tree/main/Data/factorizations fails to cover the interval [43632, 45000] without `?` marks for a couple reasons: the $T$ obtained is not always optimal and many $N$ are skipped.

plby:
I think running the integer programs to completion is not feasible for me.  I tried running the first 20 `?` for half an hour each and none have finished.

I can try some sort of decreasing number-of-factors process as you describe.

plby:
Okay, I ran the integer programs with the target number of factors decreasing by one.  As before, I cut off the run if it takes over a minute.  The results are now lower bounds.  I added a header to the file, so it now looks like this:
```
N T LP_floor IP_lower LP_upper dual_sortedness
40000 13334 39993 39997 39997.47427707 sorted
40001 13334 39992 39998 39998.58893746 sorted
40002 13334 39995 39999 39999.70460659 sorted
40003 13335 39994 40000 40000.38429752 sorted
40004 13335 39994 40001 40001.49862259 sorted
40005 13335 39995 40002 40002.61432507 sorted
40006 13336 39998 40003 40003.50201748 sorted
40007 13336 39999 40004 40004.60827169 sorted
40008 13336 39998 40005 40005.72394082 sorted
40009 13337 40001 40006 40006.42535306 sorted
40010 13337 40001 40007 40007.52185609 sorted
40011 13337 40002 40007 40008.63752522 sorted
```
The last line above is an example where the IP_lower bound is below the floor of the LP_upper bound.  The distribution of floor(LP_upper)-IP_lower is as follows:
```
   4924 0
     70 1
      7 2
```

teorth:
Thanks!  Does this now reprove $t(N) \geq N/3$ for $43632 \leq N \leq 45000$, even if it doesn't quite get the same values as in the factorizations directory?

teorth:
Here is the combined plot.  The integer programming lower bound indeed stays extremely close to the floor of the linear programming upper bound; the smoothfac lower bound loses maybe two or three factors; and the floor method loses like 10.

surplus

teorth:
Looking at the small number of green dots on the -1 line to the right of N=43632, it appears that this "impatient integer program" doesn't quite prove $t(N) \geq N/3$ for $N= 44280, 44286, 44739, 44756$, which is slightly unfortunate.  I don't know if there is a quick fix to this integer program to address these cases (I guess one also could play with increasing t above N/3, which is implicitly what is done in https://github.com/teorth/erdos-guy-selfridge/tree/main/Data/factorizations if I understand correctly?)

plby:
I'll run the program for longer on the 3 points you identified specifically, and also all 77 "stragglers".

plby:
I ran the program for longer on the 4 (not 3, oops) points you identified specifically; sorry I missed them earlier.  They all improved above the necessary threshold; in fact, they all improved to the (floor of the) LP upper bound, which was 1 or 2 up depending.
```
-44280 14760 44272 44279 44280.89900111 sorted
+44280 14760 44272 44280 44280.89900111 sorted
-44286 14762 44279 44285 44286.78535607 sorted
+44286 14762 44279 44286 44286.78535607 sorted
-44739 14913 44733 44738 44740.40857584 sorted
+44739 14913 44733 44740 44740.40857584 sorted
-44756 14919 44750 44755 44757.42691764 sorted
+44756 14919 44750 44757 44757.42691764 sorted
```
I'm also trying the other 77-4=73 points where the IP lower bound was weak.  Presumably some of them are actually strong because the LP upper bound is weak ...

teorth:
It's great that you've confirmed the previous findings that integer programming can prove $t(N) \geq N/3$ in this range!  But it makes the integer programming results more difficult to reliably replicate because we have made some ad hoc decisions about when to abandon any given integer programming attempt.  This is in contrast to, say, the vanilla greedy method or the vanilla linear programming upper bound (or floor lower bound), which have precise unambiguous implementations that can be replicated.  On the other hand, the factorizations that are produced can be reliably *verified* at least.  I think I will downgrade the integer programming method to an "approach" rather than a "method" and mention that the output and performance can depend on implementation decisions.

PR #44, opened 2025-04-30
Optimal greedy bounds

AndrewVSutherland:
I computed some optimal bounds for the original greedy bounds (using the same approach as for fastegs, get a good lower bound heuristically, get a hard upper bound by taking the largest t for which the amount left over at the point where you would be forced to use a sub-optimal cofacter still has enough bits to give N factors if things split perfectly, and then check every t in the interval in between).  Here is a table covering 10^4 to 10^6 in `tbounds_g.txt`

| $$N$$ | $$\lceil\frac{N}{3}\rceil$$ | $$t(N)\ge $$ |
|---:|---:|---:|
| $$1\times 10^4$$ | $$3334$$| $$3258$$ |
| $$2\times 10^4$$ | $$6667$$| $$6578$$ |
| $$3\times 10^4$$ | $$10000$$| $$9912$$ |
| $$4\times 10^4$$ | $$13334$$| $$13303$$ | 
| $$5\times 10^4$$ | $$16667$$| $$16667$$ | 
| $$6\times 10^4$$ | $$20000$$| $$19950$$ |
| $$7\times 10^4$$ | $$23334$$| $$23414$$ |
| $$8\times 10^4$$ | $$26667$$| $$26791$$ |
| $$9\times 10^4$$ | $$30000$$| $$30177$$ |
| $$1\times 10^5$$ | $$33334$$| $$33572$$ |
| $$2\times 10^5$$ | $$66667$$| $$67613$$ |
| $$3\times 10^5$$ | $$100000$$| $$101849$$ |
| $$4\times 10^5$$ | $$133334$$| $$135696$$ | 
| $$5\times 10^5$$ | $$166667$$| $$170405$$ | 
| $$6\times 10^5$$ | $$200000$$| $$204597$$ |
| $$7\times 10^5$$ | $$233334$$| $$238908$$ |
| $$8\times 10^5$$ | $$266667$$| $$273378$$ |
| $$9\times 10^5$$ | $$300000$$| $$307803$$ |
| $$1\times 10^6$$ | $$333334$$| $$342303$$ |

I then noticed that one can verify the entire range from 67425 to 80000 using the greedy algorithm, so I think the $$8\times 10^4$$ that currently appears in the paper on page 12 can be improved.  This PR includes a hint file `h_67425_80000_g.txt` with $$N:t$$ pairs that cover this range.

I also noticed that the greedy algorithm can prove $$t(N) >= N/3$$ for $$N$$ as small as 44716, where it can prove $$t(N) \ge 14906 = \lceil\frac{N}{3}\rceil$$.  This is lower than I would have guessed. 

But it might good for someone to independently verify these claims.


AndrewVSutherland:
I've added code that provides an option to dump factorizations and verify the results, so I'm now confident these claimes are correct.  I want to cleanup the code a bit but I'll push another commit later today.  I'm also in the process of trying to determine the complete set of integers for which the greedy algorithm can prove t(N) >= N/3 (which should include [67425,infty) and a substantial subset of [44716,67425).

But this raises the question of whether the proof we have that t(N) >= N/3 for all N >= 10^11 actually proves that the greedy algorithm achieves this.  Whle I'm confident this is true, I don't think this is immediate, which raises the question of what is the least N for which we can easily prove that the greedy algorithm will output at least N factors for some t >= N/3?

teorth:
Great question!  Unfortunately the greedy algorithm is difficult to analyze analytically for N very large.  The initial stages are relatively easy, in particular greedily clearing out all primes down to $\sqrt{N}$ is straightforward, which takes care of about $N \log 2$ of the $N$ desired factors.  With a bit more effort one can probably analyze what happens for say the first $(1-\varepsilon) N$ factors produced by the greedy algorithm for any fixed $\varepsilon$ (the answer will be some recursively defined thing somewhat similar to the Dickman and de Bruijn functions).  But that final $\varepsilon N$ is tricky.  The problem is that the initial greedy steps may accidentally exhaust all the small prime factors such as 2,3,5 and suddenly the greedy algorithm becomes very inefficient in the final stages.  One can work with various modified greedy algorithms that keep e.g., all the powers of 2 in reserve until one only has $\varepsilon N$ factors remaining, but these are somewhat ad hoc techniques and the analysis is likely to be at least as lengthy as the analysis of the modified approximation factorization method that is already in the paper.  But presumably the numerics become extremely convincing that we could safely conjecture that the vanilla greedy method should be able to certify $t(N) \geq N/3$ for all $N \geq 10^{12}$ say.

teorth:
Would you also be able to provide data on how the greedy algorithm performs to find N/3-admissible factorizations of N! for N in [40000,45000], for comparision with the linear programming methods?  Presumably they should almost always fall short of N factors, but I would be curious to see data on how far short they fall typically.

PR #51, opened 2025-05-01
greedy factor counts for N in [40000,45000] with t=ceil(N/3)

AndrewVSutherland:
The file /src/fastegs/Mbounds_40000_45000_g.txt contains lines `N:t:cnt:delta` for N in `[40000,45000]`` with t=ceil(N/3), where cnt <= M(N,t) is the number of factors the greedy algorithm produces on inputs N and t, and delta is cnt-N, which is nonnegative for 36 values of N in this range, the smallest of which is N=44716.


teorth:
Thanks!

PR #55, opened 2025-05-01
Add lower range 1e4 - 9e4 to 'lp_large_n.txt'

uhrm:
This PR adds LP bounds data for the range 10000-90000.

This is in response to the call to action in Figure 9 of the manuscript.


teorth:
Thanks!  One thing I noted on plotting is that the upper bound from Lemma 5.1 seems to be unable to go below N/3, presumably because the script to search for t stops at that level.  Is it possible to update that script to allow the optimal upper bound from Lemma 5.1 to go below N/3?

uhrm:
Thanks for spotting this. I've pushed a fix and updated the bounds.

teorth:
Thanks!  I like how the combined graph is looking now!

verylarge

PR #58, opened 2025-05-02
Add a proof that it is NOT possible to get 1/4 by rearranging only 2,3.

plby:
In the Monthly paper "Factoring Factorial n", Guy and Selfridge state that

> By similar methods we are also able to prove that, for sufficiently large $n$, $n!$ can be expressed as the product of $n$ factors, each greater than $n/4$, by rearranging powers of two and three only.

In this PR, I give a proof that the opposite is true: for all sufficiently large $N$, it is NOT possible to express $N!$ as the product of $N$ factors, each of which is at least $N/4$, by rearranging powers of two and three only.

The proof is via a (dual) linear program, which I tried to code up in a very simplistic way so that it would be easy to check that it does what it says.  (As with my previous dual linear program verification, everything is done in exact arithmetic.)  In the process of writing the program, it occurs to me that this proof IS fundamentally human-readable mathematics, but I have not transformed it into that form yet.

The idea is this: take all the numbers $1,\dotsc,N$ and divide out all the factors of $2$ and $3$.  Then try to multiply them back so that each term is at least $N/4$.  (If your strategy for doing so does not involve dividing out a factor from a number, that's no matter, because you can divide it out and multiply it right back later.  So wlog, the first step is to divide all 2s and 3s out.)

For large $N$, we can figure out what fraction of the factors (up to $o(1)$ of the fraction terms) will need to be multiplied by at least $x$ for any $3$-smooth number $x$.  Call this fraction $d_x$; in the program it is called `dist[x]`.

Let $a_x$ denote the fraction of numbers that is, in the end, in fact multiplied by $x$.  (In the paper, we have $a_\ell$ even for non-smooth $\ell$ but the situation is equivalent because you can't use non-smooth $\ell$ anyway.)  We will add up a bunch of inequalities of the form $\sum_x lhs_x a_x \le rhs$.  (We will also multiply the inequalities by a non-negative coefficient as we do so.)  The goal is to get an inequality like $\sum_x a_x \le 1-\varepsilon$ for a specific positive $\varepsilon>0$.

The inequalities are of two flavors.  The first flavor is the prime budgets: $\sum_x \nu_2(x) a_x \le 1$ and $\sum_x \nu_3(x) a_x \le 1/2$.

The second flavor is what I call "earth-moving".  There are only so many numbers available to multiply by each $x$, so the inequalities here are like $a_1 \le d_1$ and $a_1+a_2\le d_1+d_2$ and ... and $a_1 + a_2+a_3+a_4+a_6\le d_1+d_2+d_3+d_4+d_6$ and so forth.

It turns out if you add these with the right coefficients, then you get a sum of the form $\sum lhs_x < rhs_x$ where each $lhs_x \ge 1$ while $rhs_x < 1$.  The result follows.

The coefficients are as follows:
* The prime budget for $2$ has coefficient $2/27$.
* The prime budget for $3$ has coefficient $3/27 = 1/9$.
* The earth-moving constraint for $1$ has coefficient $2/27$.
* The earth-moving constraint for other "nice" numbers has coefficient $1/27$.  A number $x=2^i\cdot 3^j$ is nice if $i\le 2$ and $x\le M$ (see below).
* All other earth-moving constraints have coefficient $0$.

Implementation detail: To make the program more finite, I actually cut off the smooth numbers at some point.  Specifically, if a number ends up multiplied by $x=2^i\cdot 3^j$, I just group it with $2^{\min(i, 16)}\cdot 3^{\min(j,10)}$.  This is beneficial to the putative re-arranger in terms of the prime budget; meanwhile, I do not use the earth-moving constraints if $x > M=13122 = 2\cdot 3^8$, so the truth of those inequalities is not affected by this grouping.  In other words, the result holds even if I don't do this grouping: the grouping was just useful to me for the verification program.

The final result has $\sum_x a_x \le 13060647757/13060694016 \approx 0.999996458151$, so $\varepsilon \approx 3.5\cdot 10^{-6}$.

### DIFF
--- a/src/dnup/one_fourth.py
+++ b/src/dnup/one_fourth.py
@@ -1,13 +1,14 @@
 from fractions import Fraction
 import logging
+import math
 
-#logging.basicConfig(level=logging.INFO)
-logging.basicConfig(level=logging.DEBUG)
+logging.basicConfig(level=logging.INFO)
+#logging.basicConfig(level=logging.DEBUG)
 
 # Parameters for the verification
-MAX_TWO = 16
-MAX_THREE = 10
-M = 2 * 3**8 # 13122
+MAX_TWO = 19
+MAX_THREE = 12
+M = 2**2 * 3**9 # 78732
 
 # Generate some 3-smooth numbers, including all of them less than 4*M
 assert 2**MAX_TWO > 4*M
@@ -75,16 +76,20 @@ rhs = Fraction(0, 1)
 # The first two inequalities will be the prime budgets.  These are
 # inequalities where the lhs is of the form sum_x v_p(x) a_x.
 
-# The budget for 2 is 1.  The coefficient of this inequality is 2/27.
-coefficient = Fraction(2, 27)
+# All coefficients will have denominator 32.
+denominator = 32
+
+# The budget for 2 is 1.  The coefficient of this inequality is 2/32 =
+# 1/16.
+coefficient = Fraction(2, denominator)
 rhs += coefficient * Fraction(1, 1)
 for two in range(MAX_TWO+1):
     for three in range(MAX_THREE+1):
         x = 2**two * 3**three
         lhs[x] += coefficient * two
 
-# The budget for 3 is 1/2.  The coefficient is 3/27 = 1/9.
-coefficient = Fraction(3, 27)
+# The budget for 3 is 1/2.  The coefficient is 3/32.
+coefficient = Fraction(3, denominator)
 rhs += coefficient * Fraction(1, 2)
 for two in range(MAX_TWO+1):
     for three in range(MAX_THREE+1):
@@ -105,14 +110,14 @@ for x in smooth:
 for two in range(MAX_TWO+1):
     for three in range(MAX_THREE+1):
         x = 2**two * 3**three
-        # The coefficient is 2/27 if x=1, 1/27 if x is nice, and 0
+        # The coefficient is 2/32 if x=1, 1/32 if x is nice, and 0
         # otherwise.  Nice means that x<=M and the power of 2 is at
         # most 2.
         coefficient = 0
         if two <= 2:
-            coefficient = Fraction(1, 27)
+            coefficient = Fraction(1, denominator)
         if x == 1:
-            coefficient = Fraction(2, 27)
+            coefficient = Fraction(2, denominator)
         if x > M:
             coefficient = 0
         if coefficient == 0:
@@ -143,4 +148,58 @@ for x in smooth:
 logging.debug(f"{float(rhs)=} {rhs=}")
 assert rhs < Fraction(1, 1)
 
-print("We have proven that, for sufficiently large N, it is NOT possible to split N! into N factors, each of which is >=N/4, by only moving factors of 2 and 3.")
+# That completes the proof, but let's also compute the bound in
+# another (faster) way and check that it's equal.  Along the way we
+# also compute a bound on N when this argument takes effect.
+def sumd(k):
+    result = 0
+    for m in smooth:
+        if m < 4*k:
+            result += (Fraction(1,m)-Fraction(1,4*k)) * Fraction(1,3)
+    return result
+def sume(k):
+    result = 0
+    for m in smooth:
+        if m < 4*k:
+            result += Fraction(4,3)
+    return result
+
+c = 0
+b = 0
+b += Fraction(2, denominator) * Fraction(1, 1) # 2-budget
+b += Fraction(3, denominator) * Fraction(1, 2) # 3-budget
+for two in range(MAX_TWO+1):
+    for three in range(MAX_THREE+1):
+        x = 2**two * 3**three
+        # The coefficient is 2/32 if x=1, 1/32 if x is nice, and 0
+        # otherwise.  Nice means that x<=M and the power of 2 is at
+        # most 2.
+        coefficient = 0
+        if two <= 2:
+            coefficient = Fraction(1, denominator)
+        if x == 1:
+            coefficient = Fraction(2, denominator)
+        if x > M:
+            coefficient = 0
+        if coefficient == 0:
+            continue
+        b += coefficient * sumd(x)
+        c += coefficient * sume(x)
+assert rhs == b
+
+print("We have proven that, for sufficiently large N, it is NOT possible to split N! into N factors,")
+print("each of which is >=N/4, by only moving factors of 2 and 3.")
+print("")
+print("Specifically, it is not possible to get asymptotically more than b*N factors, where")
+print(f"b = {b} ~ {float(b)}.")
+print("")
+
+print(f"Note that b = 1 - epsilon, where epsilon ~ 1/{math.floor(1/(1-b))}.")
+print("")
+
+print("The 'sufficiently large' can be quantified.  We get an inequality 1 <= b + O_\le(c / N) for b as above and")
+print(f"c = {c} ~ {math.ceil(c)}, from which it follows that it holds for N >= c/(1-b):")
+print("")
+
+N0 = math.ceil(c/(1-b))
+print(f"N >= {N0}.")
--- a/src/dnup/one_fourth.py
+++ b/src/dnup/one_fourth.py
@@ -1,0 +1,146 @@
+from fractions import Fraction
+import logging
+
+#logging.basicConfig(level=logging.INFO)
+logging.basicConfig(level=logging.DEBUG)
+
+# Parameters for the verification
+MAX_TWO = 16
+MAX_THREE = 10
+M = 2 * 3**8 # 13122
+
+# Generate some 3-smooth numbers, including all of them less than 4*M
+assert 2**MAX_TWO > 4*M
+assert 3**MAX_THREE > 4*M
+
+smooth = []
+for two in range(MAX_TWO+1):
+    for three in range(MAX_THREE+1):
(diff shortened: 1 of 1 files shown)


plby:
Call a number $N$ *good* if $N!$ can be expressed as the product of $N$ factors, each greater than $N / 4 $, by rearranging powers of two and three only.  (And call it *bad* otherwise.)  I will use the strict "greater than" here because that's what Guy & Selfridge do, and they provide all the bad $N$ for $N\le 221$; my calculations agree exactly with theirs.

Now, count the number of good $N$ in every block of $1000$, so $0<N\le 1000$; $1000<N\le 2000$; and so forth.  I get the following data so far:
```
1000 872
2000 771
3000 776
4000 600
5000 517
6000 681
7000 315
8000 441
9000 746
10000 99
11000 256
12000 315
13000 0
14000 109
15000 51
16000 84
17000 246
18000 208
19000 2
20000 0
21000 0
22000 0
23000 0
24000 15
25000 54
26000 82
27000 136
28000 0
29000 0
30000 0
31000 0
32000 0
33000 0
34000 0
35000 0
36000 0
37000 0
38000 0
```
The oscillatory behavior is quite interesting.  It reminds me of the behavior of the excess with respect to $N/3$ analyzed elsewhere.

Guy & Selfridge state, without proof, that this function is eventually always $1000$.  I prove above that it is eventually $0$, and I really think a quite effective bound can be extracted.  The largest $N$ which was good *so far* is $N=26243$.  Maybe it's the last one?

teorth:
Interesting!  So it seems Guy and Selfridge made a mistake somewhere.  (They both passed away, so one can't ask them directly.)  Was there a particular reason for your choice of coefficients 2/27, 3/27, etc., or were they just the first ones that worked?

It might be possible to get an explicit bound on N beyond which we can show that t=N/4 is not possible with this method.  The prime budget bounds are exact for any N because $\nu_2(N!) \leq N$ and $\nu_3(N!) \leq N/2$.  The exact value of $d_1+\dots+d_k$ for a fixed N (not asymptotic) is, I think
$$ \sum_{m \leq 4k, \hbox{ smooth}} \# \{ m N/4k < n \leq N; n \hbox{ rough} \} / N$$
where "rough" means coprime to 2,3 and "smooth" means having only 2,3 as prime factors.  Using Lemma 9.1 from the paper this gives an interval arithmetic bound
$$ d_1 + \dots + d_k = \sum_{m \leq 4k, \hbox{ smooth}} ((1 - \frac{1}{4k})/3 + O_{\leq}(4/3N)).$$
The first term should hopefully correspond to what your calcultions are also giving.  If one then takes the particular linear combination of the inequalities you give, this should give a claim of the form
$$ 1 \leq 0.999996458151 + O_{\leq}( C N)$$
for an explicitly computable $C$.  This should then give an explicit threshold beyond which things break down.

plby:
The coefficients are the (dual) solution for the optimum bound for this value of $M$. I think it would be best to pick the slightly bigger $M$ with the better bound and the corresponding best coefficients there. I think they're multiples of $1/35$ or something.

plby:
My program to check $N$ one-by-one has found no more instances where the G&S statement works up to 175000, so $N=26243$ is still conjecturally the highest.  (If you replace the strict inequality with the weak $\ge N/4$, then you can get one more: $N=26244$.)

A little bit of numerology: we've been proving that $t(N)/N\ge r$ for different ratios $r$ and different ranges of $N$.  The absolute minimum ratio is $\min_N t(N)/N=1/4$, achieved only at $N=8$.  (Our results prove this.)  As such, perhaps Guy & Selfridge thought it would be nice if working with only the primes $2$ and $3$ showed this bound asymptotically.  And impressively, the proof technique does work for the weak $\ge N/4$ version for all $N<1088$; for it to then eventually always fail is potentially slightly surprising!  [The absolute maximum is $\max_N t(N)/N = 1$, achieved only at $N=1$.]

As mentioned in the abstract, the final cutoff for the ratio $1/3$ is $43632 = 432\cdot 101 = 2^4\cdot 3^3\cdot 101$.  The conjectural cutoff for the discussion in this thread is $26244 = 4\cdot 6561 = 2^2\cdot 3^8$.  Presumably the relatively large powers of $2$ and $3$ are far from accidental.

plby:
In the better bound for larger $M$, the coefficients are:
* $2/35$ for the $2$-budget,
* $3/35$ for the $3$-budget,
* $2/35$ for earth-moving constraint $1$,
* $1/35$ for other nice earth-moving constraints,
* and $0$ otherwise.

The earth-moving constraint $x=2^i\cdot 3^j$ is "nice" if $i\le 2$ and $x\le M:=236196=2^2\cdot 3^{10}$.  The bound achieved is roughly $\varepsilon\approx 1/19451$.  I'll try to see what this gives when you use precise bounds.

evan-conway:
This is somewhat out of scope, but I would be curious what happens if you also allow rearrangement of factors of 5 (or 7, 11, etc). This may be more difficult to analyze, however.

plby:
@evan-conway : I think if you use primes up to 5, you get a fraction around 0.267, but notably less than $2/7$.  I think if you use primes up to 7, you get a fraction strictly above $2/7$.  Moreover, I think it's achievable with significant slack; I'm working on working that out and adding it to the paper hopefully.

evan-conway:
Interesting! I look forward to seeing your results on adding 5 and 7. I would be curious to know what the highest prime is to achieve certain thresholds (ex: $t(N)/N)$). I imagine that the number of primes you need to be able to move would rise relatively quickly as you shift the threshold.

teorth:
Thanks for the correction!  Hopefully the better bound gives a civilized threshold of N for which the error terms 4/3N can be neglected.  (But note that each $d_1 + \dots + d_k$ has not just one of these error terms, but multiple copies, depending on how many smooth $m$ there are that are less than or equal to $4k$.  Actually one can drop the $m=4k$ term (i.e., sum over $m<4k$ rather than $m \leq 4k$) as that term is simply zero (with no error term), which may save a tiny bit in the total error term accounting (but every little bit helps).

plby:
Oops, I misread the parentheses.  I knew it was too good to be true.

For the original bound in this PR, I get the sum is $(13060647757/13060694016)+(994/27)\cdot O_{\le}(4/(3N))$, where the second coefficient is thus $3976/81\approx 49$.  Thus overall this gives that $N\ge 1.39\cdot 10^6$ is impossible.

For the better bound, I get the bound $N\ge 1.44\cdot 10^6$ instead because the coefficient is $(1937/35)\cdot (4/3)=7748/105 \approx 74$. 😞  It's honestly pretty civilized as far as it goes, but the search is only up to 400k so far.  Presumably that can be sped up by being smarter, but I will take a look at options with the bounds first.

plby:
[Correction: the bound in this PR gives $N\ge 1.39\cdot 10^7$.]

teorth:
It's great that you've managed to cover all cases now!  I think it's worth adding this as a new component of the main theorem; I'll go ahead and do this later today, and update the rearrangement section to first talk about the non-asymptotic regime (where N is finite, rather than going to infinity).

teorth:
Now that I started writing up the theory I realize I made a mistake: the upper bounds $\nu_2(N!) \leq N$, $\nu_3 \leq N/2$ go the wrong way.  What we actually need are the lower bounds
$$ \nu_2(N!) \geq N - \lceil \frac{\log N}{\log 2} \rceil \geq N (1 - \frac{\log(2N)}{N\log 2} )$$
and
$$ \nu_3(N!) \geq \frac{N}{2} - \lceil \frac{\log N}{\log 3} \rceil \geq \frac{N}{2} (1 - \frac{\log(3N)}{N \log \sqrt{3}} ).$$
For $N > 15$, $1 - \frac{\log(3N)}{N\log \sqrt{3}}$ is smaller than $1 - \frac{\log(2N)}{N\log 2}$.  In practice, what this means is that we have to multiply the weights $a_k$ by $1 - \frac{\log(3N)}{N\log \sqrt{3}}$, effectively creating an additional $\frac{\log(3N)}{N\log \sqrt{3}}$ error term to add to all the other $4/3N$ type error terms that one already has.  This shouldn't affect the calculations *too* much, but it does mean that the threshold at which we can theoretically guarantee failure of the 2,3-rearrangement method has to be increased slightly.

plby:
I think the lower bounds are needed if we're trying to prove that a rearrangement **is** possible for $N$ of a particular size (with some downset).  And I hope that we can do this with $t(N)/N\ge 2/7$ for a civilized $N$, so I hope the analysis you're doing will be useful very soon.

But this PR is about rearranging $2,3$ **not** working for $t(N)/N\ge 1/4$, so doesn't the inequality go exactly the right way here?

teorth:
Oh, you're right, never mind.  Phew!

By the way, if you have tables of what $t_{2,3}(N)$ (defined as the largest value of $t$ one can reach purely from rearranging powers of $2$ and $3$) are, this could make a nice plot, in particular comparing $t_2(N)/N, t_{2,3}(N)/N, t(N)/N$ against $3/16$ and $1/4$ (where $t_2$ is defined in the obvious fashion).

teorth:
Actually there is one further very stupid issue: the $m=4k$ term does not, in fact, vanish, but actually can supply an additional error term of 1/N to the analysis, if it turns out that $N/4k$ is an integer.  (This is a tiny bit better than the full 4/3N term, but only by a little bit.)   So one may need to make this very slight adjustment to the error calculations.

PR #59, opened 2025-05-02
finalized fastegs code

AndrewVSutherland:
This PR contains what should bel the final version of `egs.c` along with an updated README.md, updated hint files, and some updated bounds files (there are a few more to come).

This version includes implementations of both the standard greedy algorithm (supporting N <= 2^30) and the fast variant (supporting N <= 2^40), both of which should use comfortably under a gigabyte of memory.  It also supports heuristic and exhaustive searches for the largest value of t for which each of the two algorithms can prove t(N) >= t.

I've made some changes to the code for both the heuristic and exhaustive searches to make them faster.  The heuristic search now produces slightly different (usually better) values and is 2-3 times faster (it takes about 20 minutes to create a hint file up to 10^12, and I've updated the hint files to match the new algorithm.  The exhaustive search is now fully parallelized and will use as many cores as are available to speed up the search.

I've also added options that will record factorizations as they are computed, and let you to ask the algorithm to explicitly verify the factorizations after the fact (this will of course slow the algorithm down but is still feasible even for fairly large values of N), and to dump a factorization to a file if you wish.  While doing so I found a minor bug in the fast greedy algorithm that was causing it to produce bounds slightly worse than it should have, and in a few edge cases, slightly better (and potentially incorrect, although these were all well below the greedy bound so never actually false).  This bug has now been fixed.

An unfortunate side effect is that both the heuristic and exhaustive bounds for the fast greedy algorithm, including those plotted in Figure 8, are going to be slightly different (probably the difference is too small to be visible but I think we should update the table).  I'm still in the process of recomputing them and will push another commit once those computations finish.


teorth:
Thanks!  I updated the table.  (I am transferring the data to tables.py in a semi-automated fashion (I paste in the data as raw text and guide Github copilot to do the data conversion).  In retrospect I should probably have written a crude parser, but in any case the table is done and presumably the files are now stable.)

AndrewVSutherland:
Thanks, and sorry for the hassle (I promise not to change them again!).  I'm working on Section 3 of the latex file now explaining some of the algorithmic details and should have a PR ready later tonight.

PR #64, opened 2025-05-05
Section 3 updates

AndrewVSutherland:
I sharpened the complexity analysis in section 3 to show that the fast variant of the greedy algorithm can be implemented to achive an O(N^{2/3+espilon}) running time (versus the O(N) complexity of the basic greedy algorithm), and noting that our implementation is actually O(N^{3/4+epsilon}) becase we are using the Lehmer-Miller-Odlyzko algorithm to compute pi(x) (rather than the asymptotically faster but less practical analytic method)

I still need to finish Section 3.3, which I will try to get done tomorrow.


evan-conway:
Is this ready to be merged once the conflicts are resolved? If so, I can try to resolve those and merge this in.

AndrewVSutherland:
There are a few more updates pending; after doing the asymptotic analysis I realized it might be interesting to extend the verification further to get a better sense of how the performance scales. I've now gone out to 10^14, and I want to update the table with the timings, and include so information on how the hint files were generated.

I will try to get this done tomorrow (sorry for the delay, I was busy with a conference all last week and just back yesterday).

I'm going to close this PR, and tomorow I'll submit a new one that won't have merge conflicts (in this branch the only changes are local to Section 3 and I can easily merge them into the branch that has the code changes and the new data).

PR #65, opened 2025-05-05
My CRediT contribution

matthieu-rosenfeld:
I added my name, and CRediT contribution, even if the code I have written is now "obsolete" compared to the much faster version by @AndrewVSutherland.

### DIFF
--- a/LaTeX/notes.tex
+++ b/LaTeX/notes.tex
@@ -133,6 +133,10 @@
 \address{UVA Department of Mathematics, Charlottesville, VA 22903}
 \email{auj4kq@virginia.edu}
 
+\author[Rosenfeld]{Matthieu Rosenfeld}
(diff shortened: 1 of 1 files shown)


teorth:
My personal view is that it is the amount of useful input to the project that primarily determines the level of contribution, not the amount of output of the project that can be directly attributed to the contributor.  And the fact that we have independent ways to confirm the output of greedy algorithms is indeed very reassuring!

matthieu-rosenfeld:
I agree that it is better to have multiple independent implementations, and I was happy to see the ideas you used to improve the previous version. By the way, I guess that the last version of the C++ implementation could be run for up to $10^{12}$ in ~10h using ~40GB of memory. But I do not think that we get anything interesting by running it as it does not even try to provide "good" bounds on $t(n)$ anymore.

PR #66, opened 2025-05-05
Edits in LP sections

uhrm:
This PR has two commits with edits in the linear programming sections of the manuscript.

The first commit fixes two typos in proofs of Proposition (4.2) and Lemma (5.1). Please double-check before merging.

The second commit adds references to the linear programming software used in our numerical computations. The TODO in the manuscript also mentioned Mojo, but that is just the programming language used (instead of Python) and probably doesn't need to be mentioned explicitly.


teorth:
Looks good to me.  Thanks for the corrections!

PR #67, opened 2025-05-05
Add programs and data for rearranging only small primes.

plby:
This PR includes a Python program that accepts $(P,N,T)$ and generates a lp-format integer program to determine whether $t(N)\ge T$ can be demonstrated only by rearranging primes up to $P$.

It also includes a C++ program that computes $t_2(N)$ at very fast speed, like a million $N$ in a few seconds.

Finally, I've included data files for $t_2(N), t_{2,3}(N), t_{2,3,5}(N), t_{2,3,5,7}(N)$ for small $N$.


teorth:
Thanks!  They make for some nice plots, which I have added to the paper.

PR #68, opened 2025-05-07
Include a proof of t_{2,3,5,7} >= N*2/7 for sufficiently large N.

plby:
I've added a program that verifies a proof that t_{2,3,5,7} >= N*2/7 for sufficiently large N.  It uses Proposition 6.3 (Criterion for asymptotic lower bound) directly.  Sadly, this is the simplest setup I found.

The downset $\mathcal{D}$ is all $7$-smooth numbers up to $28$.  The sequence $a_\ell$ is computed using some linear program that tries to maximize some notion of slack.  The values of $a_\ell$ for $\ell\<52$ are specified directly by rounding the LP solution.

One enormous technical difficulty is that the sequence $a_\ell$ must have infinitely-many nonzero entries for this approach to work.  We use the idea of setting $a_\ell = a_{\ell/2}/2$ if $\ell\ge 52$.  In this way, the terms in (6.8) and (6.9) are manageable infinite series.

I have two major tasks for myself:
1. Estimate the $N_0$ for which this proves that "for all $N\ge N_0, t_{2,3,5,7}\ge N\cdot 2/7$".  I expect it to be decent.  Once I get some control of the estimates, I could see whether I can modify the weights to improve it.
2. Write this up in Section 6.

### DIFF
--- a/src/dnup/two_sevenths.py
+++ b/src/dnup/two_sevenths.py
@@ -1,0 +1,127 @@
+# This program is supposed to verify that t_{2,3,5,7} >= N*2/7 for
+# sufficiently large N.  It follows the recipe of Proposition 6.7.
+
+from fractions import Fraction
+
+def valuation(N: int, P: int) -> int:
+    """
+    Compute the P-adic valuation of N.
+
+    The P-adic valuation v_P(N) is the largest exponent k such that P**k divides N.
+    """
+    result = 0
+    while (N%P) == 0:
+        N //= P
+        result += 1
+    return result
+
+# This is the asymptotic fraction we are targeting.
+alpha = Fraction(2,7)
+
+# This is our set \mathcal{D}, the 7-smooth numbers up to 28.
+primes = [2,3,5,7]
+D = [1,2,3,4,5,6,7,8,9,10,12,14,15,16,18,20,21,24,25,27,28]
+
+# This is the initial segment of the infinite sequence of non-negative
+# real numbers a_\ell.
+a = {
+    2: Fraction(145686,1000000),
+    3: Fraction(10165,100000),
+    4: Fraction(721964,10000000),
+    5: Fraction(592374,10000000),
+    6: Fraction(512691,10000000),
+    7: Fraction(411295,10000000),
+    8: Fraction(351711,10000000),
+    9: Fraction(30096,1000000),
+    10: Fraction(240768,10000000),
+    12: Fraction(361152,10000000),
+    14: Fraction(257966,10000000),
+    15: Fraction(103186,10000000),
+    16: Fraction(902881,100000000),
+    18: Fraction(15048,1000000),
+    20: Fraction(120384,10000000),
+    21: Fraction(515932,100000000),
+    24: Fraction(128983,10000000),
+    25: Fraction(361152,100000000),
+    27: Fraction(642048,100000000),
+    30: Fraction(802561,100000000),
+    32: Fraction(334431,100000000),
+    36: Fraction(86941,10000000),
+    40: Fraction(497602,100000000),
+    42: Fraction(362285,100000000),
+    48: Fraction(494001,100000000),
+    50: Fraction(331489,100000000),
+    }
+
+# As an infinite sequence, it continues with a[x]=a[x/2]/2 if x/2 >=
+# repeat_from.  For example, a[52] = a[26]/2.  As such, we will have
+# to sum infinite series below (manually).
+repeat_from = 26
+
+# Compute sigma_{d,D}
+sigma = dict()
+for d in D:
+    sigma[d] = Fraction(1,1)
+    # Compute P_+(d):
+    Pplusd = 1
+    for p in primes:
+        if (d % p) == 0:
+            Pplusd = p
+    # Defining product for sigma_{d,D}
+    for p in primes:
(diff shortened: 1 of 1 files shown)


plby:
One thing I think I would like to do is use a different proof of Proposition 6.3.

I think it should be possible to use the modified sequence $a_\ell' = \lceil a_\ell N \rceil/N$ for $\ell\le N$ (and $a_\ell'=0$ for $\ell>N$).  The condition (6.3) / (6.9) automatically improves because $a$ only got bigger.  Meanwhile, by the easy bound $\lceil x\rceil< x+1$, the LHS of condition (6.2) / (6.8) got bigger by at most $(\sum_{\ell\in S}^N \nu_p(\ell))/N$.

I feel like that last thing isn't that big, like a few powers of $\log(N)$.

teorth:
Neat!  If you could supply some data for t_{2,3,5,7} then this would complement the other plots nicely.

I had trouble working with weights a_ell that vanishes for large N because (6.3)/(6.9) would fail for ell large enough.  No matter what one does, there are always going to be some very numbers that need to be multiplied by a large \ell to get them above t.  There is considerable freedom in what weight a_\ell to assign for very large \ell but something like 1/sqrt(\ell) seems to be more than enough while still staying well under budget for large \ell, so that's what I went with.

plby:
$t_{2,3,5,7}$ is in https://github.com/teorth/erdos-guy-selfridge/blob/9a69c041d1832582867b9ff828a58764953a8809/Data/rearrange/up_to_7.txt though it's not computed quite as far as the others.

plby:
I think I succeeded getting a decent estimate with the ceiling idea?  My estimate for my latest code is that it starts working for $N\ge 2\cdot 10^6$ or even somewhat lower.  That would allow two things:
1. Combined with the greedy computations, it proves Conjecture 1 without the "modified approximate factorization" in Section 8.
2. Or, after we check a relatively small finite range, I think even more is true: $t_{2,3,5,7}(N) \ge \lfloor 2N/7\rfloor$ for all $N\ne 56$.  That is, Conjecture 1 is true even if you only rearrange primes up to 7.

plby:
Regarding the proof of Proposition 6.3: (I think) I understand the argument in the paper with the $1/\sqrt{\ell}$, but I don't quite understand your comment here.  I agree that with the way things are set up, you need $a_\ell>0$ for large $\ell$ in (6.9), but I was just talking about the reduction step *from* (6.9) *to* (6.3).  That is, we have the asymptotic (6.9) and we're trying to get (6.3) for an actual $N$.  In (6.3), you can have $a_{\ell}$ vanish for large $\ell$, right?  (Yours vanish eventually when $N/\sqrt{\ell}<1$, right?)  It seems to me that the RHS of (6.3) is 0 when $\ell>\alpha N$, so you may as well have $a_\ell=0$ then.

teorth:
Ah, I see what you are saying, (6.2) would be fine, but (6.3) becomes tricky here because by zeroing out the $a_\ell$ for $\ell > N$ you could shrink the LHS of (6.3) too much.  That was why I had to insert in the 1/\sqrt{\ell} factor, to ensure that there was still enough weight at large \ell to not have (6.3) fall apart when \ell got close to N.

plby:
But I think maybe one further miscommunication?: I'm taking a *ceiling*, not a floor. So my LHS of (6.3) only gets bigger.

plby:
Okay, I think I finally understand your comment. The LHS has a tail that includes big $\ell$ in the asymptotic version that I'm zeroing out.

Sorry for taking so long to get the point ...

plby:
I think for my particular setup, it should all work out, because my tail decays exponentially and I can figure out the *truncated* LHS sums of (6.3) pretty exactly. I'll need to be more careful, but I think the actual effect is very small.

plby:
Recent changes (committed to main):
* I added values of $t_{2,3,5,7}(N)$ for all $N\le 50000$ in https://github.com/teorth/erdos-guy-selfridge/commit/5a3f953a8d6beea4a4b5bb8b8aea01f3a3a7ac54
* I added a "ladder" of values of $t_{2,3,5,7}(N)$ for $N$ from $57$ up to past $12$ million.  There are only $635$ values.  The sequence is the sparsest possible that shows that for all $N$ in the range, $t_{2,3,5,7}(N)\ge\lfloor 2N/7\rfloor$.  In this sequence, the ratio $t_{2,3,5,7}(N)/N$ looks like it's approaching $0.290582(1)$ or so.

plby:
I have done all the estimates by using the approach of ceilings and truncations.  I was able to control the tails using the exponential decay, but I needed to go out a bit further to make it all work: I define $a_\ell' = \lceil a_\ell N\rceil/N$ for $\ell < 2^7 N$ and $0$ otherwise.  The bound I ended up getting was it all works for $N\ge 9\cdot 10^6$.  And that works well with the "ladder" verification up to $12\cdot 10^6$ mentioned in my previous comment.  Yay!

teorth:
That's great! Do you feel like you can update the paper to incorporate this?  Or I can try to take a look at it, but perhaps since it's still fresh in your mind it would be easier for you to make the first draft.

plby:
@teorth : Yes, I wrote up the proof in the separate PR https://github.com/teorth/erdos-guy-selfridge/pull/69

PR #70, opened 2025-05-12
update egs code and timings

AndrewVSutherland:
Extends supported range of the fast variant of the greedy algorithm to 10^14 and updates timings and hint files


AndrewVSutherland:
I will create a separate PR for the edits to the latex file (including updates to the tables related to the timings).

PR #72, opened 2025-05-14
Reorder the rows of Table 1.

plby:
I already committed a change that attempted to make Table 1 prettier but without editorializing.

In this PR, I wonder whether it would be nice to re-order the rows.  The result looks like this now:

Screenshot from 2025-05-13 21-55-18

What do you think?


plby:
This isn't related to the re-ordering, but my reasoning for removing the lower bound $43632$ is that part (iii) says not only that something is true starting then, but also that the threshold is best possible.  So at the very least it includes $43631$ as well. But also I think it looks very clear now that the entire range of $N$ is verified for each part [except (iv)].

evan-conway:
This looks great to me! When skimming through the paper before, I'd thought that the table was a bit difficult to parse, and I definitely think this makes it more readable.

PR #73, opened 2025-05-15
A few more fastegs updates

AndrewVSutherland:
This PR fixes a minor bug in the code producing factorization certificates (the factorizations were always correct, the bug was in the generation of the certificate), and it generates better hint files.  The previous version was not making full use of the fact that monotonicity of $$t(N)$$ means that any bound of the form $$t(N)\ge t \ge N/3$$ implies $$t(M)\ge t\ge M/3$$ for all $$M\in [N,3t]$$ (if anyone disagrees with this claim please say so, I noticed that the code was using the more pessimistic range $$[N,N+t-N/3]$$ and I honestly have no memory of why I did this, and now the hint files are all shorter by a factor of 3).

The README.md file in src/fastegs has also been updated to reflect all the changes over the past few weeks.


AndrewVSutherland:
I'm going to go ahead and merge this PR (which is localized to src/fastegs) so that my upcoming PR with changes to the latex file will be easier to review.

PR #74, opened 2025-05-15
Finished Section 3

AndrewVSutherland:
This PR contains a more or less final version of Section 3 (I still want to proof-read it again) that is ready to review.

I moved the table with timings for the greedy algorithm (now Table 2) up to the end of Section 3.  It includes references to linear programming lower bounds defined in Section 4, but this is explained in the caption and I think the table makes for a nice transition between Sections 3 and 4 (but I am happy to move it back if others disagree).


LaTeX/notes.tex
```diff
@@ -563,13 +563,13 @@ \section{Greedy algorithms}\label{greedy-sec}
...
+Let $x_0=\lfloor N/M\rfloor \approx N^{\nicefrac{1}{2}}$, and partition the interval $(M,N]$ into subintervals $(x_k,x_{k+1})$ on which the step functions $f_t(x)\coloneqq\lceil t/x\rceil$ and $g_N(x)\coloneqq\lfloor N/x\rfloor$ are both constant, such that $x_r=N$, and $x_k$ is a point of discontinuity for at least one of $f_t(x)$ and $g_N(x)$ for $0<k<r$.  Under our assumption that the greedy algorithm uses optimal cofacters $m_i=\lceil t/p_i\rceil=f_t(p_i)$ for each prime $p_i>M$, we will have $g_N(x_k)$ factors $m_ip_i\ge t$ with $m_i=f_t(x_{k+1})$ for each prime $p_i\in (x_k,x_{k+1}]$, and we can compute the number of factors produced by the greedy algorithm that are divisible by a prime $p>M$ as
```
teorth, line 566: "cofacters" ->"cofactors"

AndrewVSutherland: fixed

AndrewVSutherland:
One question I had about (3.2) is whether it would be easy to prove that this quantity is always nonnegative (provided M > sqrt(N) with p < M,  I think this is clear asymptotically but we would want an explicit statement).  We don't need this, but if it were easy to prove it would simplify the exposition slightly.

teorth:
I think it is in principle doable to establish that (3.2) holds for all p < M (with maybe a small number of exceptions for very small N) but it would be a pain - one would have to do some calculations with explicit forms of Mertens theorem, and to also keep track of lower order terms arising from fractional part roundoffs etc.  Perhaps we could supply some numerical evidence of non-negativity (e.g., setting p equal to the closest prime to $N^\theta$ for some selected $0 \leq \theta \leq 1/2$ and plotting the two terms in the RHS (3.2) multiplied by $p-1$ plotted against $\log N$ to visibly see the convergence to $1$ and $\log 2$, and just assert as a conjecture that (3.2) is non-negative for N bigger than whatever it needs to be (e.g., 10).

AndrewVSutherland:
Okay, I will take a stab at adding the plot you suggest and making the conjecture explicit.  I'm going to close this PR to fix the merge conflict and resubmit.

PR #81, opened 2025-05-25
Proof-reading updates to section 3 (plus fixes for a few latex warnings)

AndrewVSutherland:
I've fixed a number of (non-mathematical) typos in Section 3 that I caught while proof-reading, and I added Remark 3.1 to say a bit more about the heuristic that (3.2) is always non-negative.  I've verified this computationally for 10 <= N <= 10^8 and all primes p <= sqrt(N) using t = ceil(N/3) (arguably the worst case that is relevant to us), but it wuld be expensive to go much further than this, and I'm not sure a plot is useful (you really need to check all p up to sqrt(N), not a fixed p, because the p close to sqrt(N) are actually the onese where (3.2) is closest to being negative, and the minimal valuation that arises is not monotonic, so plotting (3.2) for a sample of p and N wouldn't be particularly convincing.

But as noted in Remark 3.1, all that is relevant to the proof of Theorem 1.3(iii) is that (3.2) is nonnegative in all the cases that where we used the greedy algorithm to prove a bound, and this is true (the algorithm checks and will produce an assert failure if this is not the case), and for the asymptotic claims on the running time (which include an N^epsilon fudge factor), even if (3.2) is not always true, one could have the algorithm increase M until it is and I think it is clear that some M=O(N^(1/2+epsilon)) will work asymptotically (the hard part is making all the bounds explicit).

With this PR I have now finished everything that was left on my todo list.  The only thing I could might like to see (but this is not at all essential) is extending Figure 10 to 10^14; if someone knows how to easily regenerate this plot (using data from  `src/fastegs/tbounds_heuristic_fast_1e4_1e14.txt` that would be great (if not I can take a look at doing this next week).


teorth:
Looks great!

PR #84, opened 2025-05-31
switch to 10pt font and tweak layout to reduce page count

AndrewVSutherland:
This PR changes the font size from 12pt to 10pt and reduces parskip from 5mm to 2.5mm, yielding a page count of 62 (versus 76).

I made some small changes to the introduction to improve the layout, most notably in the first sentence.  I also changed 1/\alpha x to \nicefrac{1}{\alpha x} in a few places to make it visually clearer that the x is in the denominator.

This PR should be viewed as optional, I think it's okay to stick with the larger font size if people prefer that.


evan-conway:
At least on my computer, this does not seem to reduce readability, so I'd be in favor of this change. If we need to shorten things further, I suspect looking at the graphs and tables would be a good next step, though that might not save that much space. In particular: (noting for the future)
- Figure 1 and Figure 2 could likely be merged
- Figure 6 and Figure 7 could likely be merged
- Table 2 and Table 3 could be shortened
- Some of Figures 14, 15, 16, and 17 could be cut or merged

PR #89, opened 2025-06-06
add 2/7 support to egs.c and corresponding hint file for 210-67425

AndrewVSutherland:


AndrewVSutherland:
This PR extends egs.c so it can also be used to verify the 2/7 conjecture (for N >= 210), and includes a hintfile to do this for N in [210,67425].

PR #98, opened 2026-04-02
Section 8 and 11 Edits

evan-conway:
This PR addresses most of the reviews for sections 8 and 11.

There are two fixes that I would appreciate someone else taking a look at:

- Reviewer X commented _"p. 56, (5.28): Should the second superscript also be (2)?"_. Based on the context I think this is correct, but would appreciate if someone double checked, since I'm not familiar with the math in this section.
- Reviewer X commented _"p. 52, (8.6): It appears that the multiset B (1) is defined only on the following page."_ and _"p. 52: Where was some of the notation first introduced? (For instance, K4.5 in (2.6)). This would help the reader."_. I'm not sure how best to address these issues. In particular, the first comment is a bit thorny since B^(1) seems to be used in the proposition statement, but then defined in the proof of said proposition.


teorth:
The referee is correct on both points.  I'll merge this PR and work on addressing the second point.

PR #104, opened 2026-05-15
[SYNTAX]: Correct Lemma 2.2 (missing parameter "K")

benediktjohannes:
I think that the parameter "K" was missing here and is necessary to be added in order to calculate Lemma 2.2 (because I don't see anything passing "K" and no global variable), but I haven't tested it, so please correct me if I'm mistaken.

Contributed by Benedikt Johannes


Chessing234:
The #104 signature change left evaluate() calling delta2_upper with the old arity — opened #111 to pass K through.

PR #105, opened 2026-05-16
[PAPER]: Correct mathematical inequality (≤ → ≥) in lower bound and update author section

benediktjohannes:
I have identified and corrected a mathematical inconsistency in the paper, specifically involving an incorrect inequality direction in a lower bound (≤ was corrected to ≥).

In addition, I have updated the authors section in notes.tex accordingly.

Unfortunately, I currently do not have the necessary tooling to regenerate the final PDF. I have therefore installed a LaTeX build platform locally and attempted to compile the document, and it appears to work correctly. However, the resulting PDF is approximately +102 KB larger than the previous version, which suggests that we may be using a different build tool or configuration. This difference might also simply be due to the updated author information, since authors are displayed on every page (or every second page), which could plausibly account for the increase in file size. For consistency, it might therefore be preferable to use your established build setup for generating the final PDF.

Moreover, the updated PDF should also include the changes from the previous revision, in particular the fixes provided by Maximilian in pull request 102.

Could you please regenerate the PDF including both the latest changes and pull request 102, and upload the updated version to arXiv (https://arxiv.org/abs/2503.20170)?

If applicable, it might also be necessary to update the authors information in the arXiv submission metadata (e.g., abstract/author fields), in case this is not fully derived from the LaTeX source.


teorth (approved):

teorth:
While your corrections were appreciated, please do not add yourself as an author to the paper unsolicited.

benediktjohannes:
> While your corrections were appreciated, please do not add yourself as an author to the paper unsolicited.

Ok, I‘m sorry, I thought that all GitHub contributors should be included, so I added Maximilian Nordler and me, but I see your point.

Should I open a new pull request and remove the entry again? I‘m sorry for the inconveniences.

benediktjohannes:
I just read in the paper that the authors on GitHub are (according to CRediT https://credit.niso.org/) listed within the paper (around page 6-8) and they were also listed on top as authors and because it wasn’t updated for a while, I just thought that every contributor (according to the role on CRediT https://credit.niso.org/) should be added.

But I definitely understand if you don’t want this to be done that way. So I can (of course) create a new PR to undo this again. I‘m sorry.

Chessing234:
Re-applied just the bv-lower direction fix from here (without the author-list edits) in #109.

PR #106, opened 2026-05-17
[PAPER]: Correct logarithmic scaling

benediktjohannes:
When re-reading the paper, I think I may have found another small mistake. I'm pretty sure that we may have missed a `\log()` around `x/L`, unless I'm overlooking something here.

I've updated the PDF as well with the fix.

And by the way: thank you again for merging pull request #105 👍

I'm not 100% sure whether you planned to include the fixes from pull requests #102 and #105 later on in a future update, or whether you maybe wanted to wait until a few more issues had been collected first (which I think might now perhaps already be the case with #102, #105, and this one combined), or whether you simply haven't gotten around to it yet (which is of course completely understandable, so absolutely no stress at all). I just wanted to mention it because I thought there was a small chance that the update may simply have been forgotten unintentionally, and I wanted to make sure it doesn't get missed.


Chessing234:
Re-applied the log fix in #110 — looks like #107 reverted it on main.

